PR #2, opened 2021-12-01
Allow configuration of chart directory when releasing

yannick-roeder:


helm-lint/action.yaml
```diff
@@ -22,5 +22,10 @@ runs:
       uses: helm/chart-testing-action@v2.0.1
 
     - name: Run chart-testing (lint)
-      run: ct lint --config ${{ inputs.lintConfig }}
+      run: |
+        if [[ "${GITHUB_REF#refs/heads/}" == "main" || "${GITHUB_REF#refs/heads/}" == "master" ]]; then
+          ct lint --config ${{ inputs.lintConfig }} --since HEAD^
+        else
+          ct lint --config ${{ inputs.lintConfig }} --since HEAD
+        fi
```
yannick-roeder, line 30: The default commit to compare against is HEAD on the default branch. This allows us to detect changed helm charts on master or main as well by comparing to the previous commit

philipp94831: Why not lint the whole thing?

yannick-roeder: Whenever there are changes detected the whole helm chart is linted, but when there are no changes no linting takes place. Without this change linting on the default branch is not working, because it always compares the current commit to HEAD by default (which is the same commit).

I thought about linting every time someone pushes without checking for chart changes. The action does not support that yet but there is an issue for that (see https://github.com/helm/chart-testing-action/issues/49). The main reason for detecting changes first is to save resources when testing it against a mini k8s cluster inside the CI. We are not using that part of the action yet, but we could in the future

philipp94831 (approved):

PR #9, opened 2022-02-28
Add helm release workflow and refactor helm lint action

yannick-roeder:
Closes https://github.com/bakdata/ci-templates/issues/4


.github/workflows/helm-lint.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  helm-lint:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ inputs.ref }}
+          fetch-depth: 2
```
raminqaf, line 31: why a fetch-depth of 2?

.github/workflows/helm-lint.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: "${{ inputs.helm-version }}"
```
raminqaf, line 36: wouldn't it make sense to move the `Check out repository` and the `Set up Helm` step to the composite action of `helm-lint` and add two additional inputs (`ref` and `version`) to it? Since both of these steps will be repeated each time you want to lint helm charts.

Then you will only call one step under the `helm-lint` job:
```yaml
      - name: Lint helm chart
        uses: bakdata/ci-templates/actions/helm-lint@main
        with:
          ref: ${{ inputs.ref }}
          version: version: "${{ inputs.helm-version }}"
          lintConfig: "${{ inputs.lint-config-path }}"
```

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: "${{ inputs.helm-version }}"
+
+      - name: Lint helm chart
+        uses: bakdata/ci-templates/actions/helm-lint@feature/helm-release-workflow
+        with:
+          lint-config-path: "${{ inputs.lint-config-path }}"
```
raminqaf, line 69: See above

.github/workflows/helm-lint.yaml
raminqaf: I don't see a point in having `helm-lint` as a reusable workflow. Because in that case it should be used in as a job. I think it would make sense to have `helm-lint` only as a composite action for using it as one of the steps in a job. WDYT?

.github/workflows/helm-lint.yaml
yannick-roeder: The lint action checks helm chart changes before linting. For that it compares the current helm chart with the helm chart in the previous commit

.github/workflows/helm-lint.yaml
yannick-roeder: You are right. I thought composite actions only work inside the same repository or when cloned (which is only true for private repos). In this case we can just call it from the other repo. I also need to change the README.md in the rood directory then

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,114 @@
...
+      
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "${{ inputs.python-version }}"
+      
+      - name: Bump version
+        if: success()
+        id: release-tag
+        uses: bakdata/ci-templates/actions/bump-version@feature/helm-release-workflow
```
yannick-roeder, line 94: TODO

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          release-type: ${{ inputs.release-type }}
+      
+      - name: Commit and push bumpversion.cfg file
+        if: success()
+        uses: bakdata/ci-templates/actions/commit-and-push@feature/helm-release-workflow
```
yannick-roeder, line 100: TODO

.github/workflows/README.md
```diff
@@ -1,38 +1,177 @@
-# python-poetry-release
+# Workflow Templates
+The following workflows can be found here:
+* [Helm Lint](https://github.com/bakdata/ci-templates/tree/feature/helm-release-workflow/.github/workflows#helm-lint)
+* [Helm Release](https://github.com/bakdata/ci-templates/tree/feature/helm-release-workflow/.github/workflows#helm-release)
+* [Python Poetry Release](https://github.com/bakdata/ci-templates/tree/feature/helm-release-workflow/.github/workflows#python-poetry-release)
```
yannick-roeder, line 5: TODO

.github/workflows/helm-lint.yaml
yannick-roeder: If I do not copy the actions into the calling repository anymore, then I encounter a problem with the lint configuration file, because the calling action does not seem to be searching for files inside the calling repository. I will try to find a solution for this issue

yannick-roeder: Nevermind, seems to work now

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+        description: "The old tag."
+        value: ${{ jobs.create-release.outputs.old-tag }}
+
+jobs:
+  helm-lint:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Lint helm chart
+        uses: bakdata/ci-templates/actions/helm-lint@feature/helm-release-workflow
```
yannick-roeder, line 56: TODO

.github/workflows/helm-lint.yaml
raminqaf: Okay good to know maybe add this as a comment in the file

actions/helm-lint/action.yaml
```diff
@@ -1,31 +1,45 @@
...
       uses: helm/chart-testing-action@v2.0.1
 
     - name: Run chart-testing (lint)
       run: |
         if [[ "${GITHUB_REF#refs/heads/}" == "main" || "${GITHUB_REF#refs/heads/}" == "master" ]]; then
-          ct lint --config ${{ inputs.lintConfig }} --since HEAD^
+          ct lint --config ${{ inputs.lint-config-path }} --since HEAD^
         else
-          ct lint --config ${{ inputs.lintConfig }} --since HEAD
+          ct lint --config ${{ inputs.lint-config-path }} --since HEAD
```
torbsto, line 43: Why since head? Don't we want to compare to main/master on a feature branch?

yannick-roeder: Yes you are right, I don't know what my thought process was when I merged this in the past. But it should be pointing to the head of the default branch

torbsto: As far as I can tell, this is the default for `--since`

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          ref: ${{ inputs.ref }}
+      
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "${{ inputs.python-version }}"
+      
+      - name: Bump version
+        if: success()
```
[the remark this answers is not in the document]
yannick-roeder: I think I had this discussion before in another PR, but I cannot find it anymore. I will just try what happens

actions/helm-lint/action.yaml
yannick-roeder: In the following link it says, that the default is be `HEAD`:
https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md

.github/workflows/helm-release.yaml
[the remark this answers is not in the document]
yannick-roeder: Yes seems to be default

actions/helm-lint/action.yaml
```diff
@@ -1,31 +1,46 @@
...
   steps:
-    - name: Set up Helm
-      uses: azure/setup-helm@v1
+    - name: Check out repository
+      uses: actions/checkout@v2
       with:
-        version: "v3.4.0"
+        ref: "${{ inputs.ref }}"
+        # fetch-depth of 2 needed for comparison of helm charts between commits while linting
+        fetch-depth: 2
```
torbsto, line 28: Does this still work? I think this has to be 0 now that we compare between branches

torbsto: Let's say the CI runs for commit E:

```
master: A --> B
               \
      feature:  C --> D --> E
```
We have to fetch the changes since B, i.e. looking 3 commits back.  We cannot anticipate the number of commits on a feature-branch, therefore this steps requires the full history ([= 0](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)).

actions/helm-lint/action.yaml
torbsto: Actually, I think your thought process was absolutely correct. Looking at the docs, the target-branch is master (and I guess they also support main). It should therefore work for a feature-branch, because it compares to the HEAD of main, which is exactly what we want.

raminqaf:
LGTM!

raminqaf (approved):

PR #12, opened 2022-03-11
Add helm deploy and destroy templates for GKE

yannick-roeder:


actions/helm-deploy/action.yaml
```diff
@@ -0,0 +1,92 @@
...
+          helm repo update
+        else
+          echo "No repository specified; Skipping step"
+        fi
+      env:
+        REPOSITORY_NAME: ${{ inputs.repository-name }}
+        REPOSITORY_URL: ${{ inputs.repository-url }}
+
+    - name: Deploy helm chart
+      shell: bash
```
philipp94831, line 86: Isn't there already an existing action for this? Azure has something, would expect the same for GH actions https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/helm-deploy?view=azure-devops

yannick-roeder: Yes, there is something like this: https://github.com/marketplace/actions/helm-deploy

This action is more specific because we have a setup process for GKE. I will test if it is possible to use the more general actions

actions/helm-deploy/action.yaml
```diff
@@ -0,0 +1,92 @@
...
+    default: "v3.8.1"
+    required: false 
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Setup environment
+      uses: bakdata/ci-templates/actions/helm-setup@feature/helm-deploy-destroy-workflows
```
yannick-roeder, line 60: TODO

actions/helm-deploy/action.yaml
```diff
@@ -0,0 +1,92 @@
...
+    required: true
+  gke-project:
+    description: "GKE project id for authentication"
+    required: true
+  gke-region:
+    description: "GKE region for authentication"
+    required: true
+  gke-cluster:
+    description: "GKE cluster for authentication"
+    required: true
```
yannick-roeder, line 15: I think one thing I can improve here is allowing Kubernetes authentication mainly via kubeconfig file instead of restricting the action to GKE only

actions/helm-deploy/action.yaml
yannick-roeder: There are sadly no official Github actions for helm deploy and destroy (other official helm actions exist, e.g. https://github.com/marketplace/actions/helm-chart-releaser). From all custom helm deploy and destroy actions on the market place, the mentioned one above (https://github.com/marketplace/actions/helm-deploy) looks most suitable, but I don't think it is unreasonable to use our own.

actions/helm-deploy/action.yaml
philipp94831: Let's do that in a follow-up

actions/helm-deploy/action.yaml
```diff
@@ -0,0 +1,92 @@
...
+      env:
+        REPOSITORY_NAME: ${{ inputs.repository-name }}
+        REPOSITORY_URL: ${{ inputs.repository-url }}
+
+    - name: Deploy helm chart
+      shell: bash
+      run: |
+        helm upgrade --debug --install --wait --timeout 1200s --force \
+          --version ${{ inputs.chart-version }} \
+          --values ${{ inputs.values-yaml }} \
```
philipp94831, line 90: What if we have multiple values? Can we use an array parameter?

actions/helm-deploy/README.md
```diff
@@ -0,0 +1,50 @@
...
+      with:
+        gke-service-account: ${{ secrets.GKE_SERVICE_ACCOUNT }}
+        gke-project: "my-awesome-project"
+        gke-region: "us-west1"
+        gke-cluster: "my-awesome-cluster"
+        release-name: "my-release"
+        namespace: "my-namespace"
+        chart: "foo/bar" # Installs the chart 'bar' from the repository called 'foo'
+        chart-version: "1.0.0"
+        values-yaml: "bar/values.yaml"
```
philipp94831, line 42: update docs

yannick-roeder: Done

actions/helm-deploy/action.yaml
yannick-roeder: Now the chart allows a string for a single helm values file (like before) or additionally a json array like this: `value-files: ["foo/values.yaml", "foo/values-1.yaml"]` for multiple value files

actions/helm-deploy/action.yaml
yannick-roeder: I created an issue: https://github.com/bakdata/ci-templates/issues/14

philipp94831 (approved):

actions/helm-deploy/README.md
```diff
@@ -0,0 +1,50 @@
...
+  - name: Deploy helm chart
+      uses: bakdata/ci-templates/actions/helm-deploy@main
+      with:
+        gke-service-account: ${{ secrets.GKE_SERVICE_ACCOUNT }}
+        gke-project: "my-awesome-project"
+        gke-region: "us-west1"
+        gke-cluster: "my-awesome-cluster"
+        release-name: "my-release"
+        namespace: "my-namespace"
+        chart: 'foo/bar"' # Installs the chart 'bar' from the repository called 'foo'
```
disrupted, line 40: is there an additional quote?

disrupted (approved):

PR #13, opened 2022-03-16
Set up pipeline for linting actions & workflows

disrupted:
Run actionlint and report problems as line comments.

example
problem matcher annotation example

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -16,5 +16,5 @@ jobs:
       - name: Check workflow files
         run: |
             echo "::add-matcher::.github/actionlint-matcher.json"
-            bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+            bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/v1.6.10/scripts/download-actionlint.bash)
             ./actionlint -color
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,20 @@
...
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check workflow files
+        run: |
+            echo "::add-matcher::.github/actionlint-matcher.json"
+            bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
```
yannick-roeder, line 19: Please pin it to the newest version (`v1.6.10`) instead of using main

disrupted: looks like the latest version is not compatible with our workflow `outputs` (https://github.com/bakdata/ci-templates/runs/5625138460)
it's fixed in main https://github.com/rhysd/actionlint/commit/77d8de26cbe771ed04028ff9da3b147deb39e187

yannick-roeder (approved):

.github/workflows/lint.yaml
disrupted: I've pinned it to this commit. When the next version is released we can pin the tag again.

Issue #16, opened 2022-03-23
Pin ci runner image version


yannick-roeder:

yannick-roeder:
So far not required for GitHub as there are a lot of actions that allow install

PR #17, opened 2022-05-03
Add kfilt to helm-gke-setup action

yannick-roeder:


actions/helm-gke-setup/action.yaml
```diff
@@ -53,6 +53,12 @@ runs:
       with:
         version: ${{ inputs.kubectl-version }}
 
+    - name: Install kfilt
+      shell: bash
+      run: |
+        curl -L "https://github.com/ryane/kfilt/releases/download/v0.0.7/kfilt_0.0.7_linux_amd64" -o /usr/local/bin/kfilt
```
philipp94831, line 59: Why not use a docker image that has kfilt available?

yannick-roeder: Yes, building a docker image would be the next step. I would prefer to move this into this issue: https://github.com/bakdata/ci-templates/issues/16 
Is that fine?

By moving it, I can first finish building the docker release ci template (see this branch: https://github.com/bakdata/ci-templates/blob/feature/docker-release-workflow/.github/workflows/docker-release.yaml) and build out proper ci-images here: https://github.com/bakdata/ci-images/tree/main

philipp94831: Please update the issue and leave a not on what to install. Also leave a TODO here to remove the installation

philipp94831 (approved):

actions/helm-gke-setup/action.yaml
yannick-roeder: Done

PR #20, opened 2022-06-22
Migrate to native Poetry caching

disrupted:
- Update to `setup-python@v4` action and migrate to native caching
- Bump Poetry version
- Update `python-setup-poetry` references in other composite actions for new release tag

tested here https://github.com/bakdata/streams-explorer/runs/7007918876

### DIFF
--- a/actions/python-poetry-bump-version/README.md
+++ b/actions/python-poetry-bump-version/README.md
@@ -7,37 +7,18 @@ in the `pyproject.toml` file. It will output the old and new bumped versions.
 
 This action uses another composite action listed below:
 
-* [python-setup-poetry](https://github.com/bakdata/ci-templates/tree/main/actions/python-setup-poetry)
+- [python-setup-poetry](https://github.com/bakdata/ci-templates/tree/main/actions/python-setup-poetry)
 
 ## Input Parameters
 
-| Name              | Required | Default Value |  Type   | Description                                                                                                                        |
-|-------------------|:--------:|:-------------:|:-------:|------------------------------------------------------------------------------------------------------------------------------------|
-| release-type      |    ✅     |       -       | string  | Scope of the release, see the official [documentation of poetry](https://python-poetry.org/docs/cli/#version) for possible values  |
-| python-version    |    ❌     |     3.10      | number  | The python version for setting up poetry.                                                                                          |
-| poetry-version    |    ❌     |    1.1.12     | number  | The poetry version to be installed.                                                                                                |
-| working-directory |    ❌     |       .       | The root directory of the poetry project. |
+| Name              | Required | Default Value |                   Type                    | Description                                                                                                                       |
+| ----------------- | :------: | :-----------: | :---------------------------------------: | --------------------------------------------------------------------------------------------------------------------------------- |
+| release-type      |    ✅     |       -       |                  string                   | Scope of the release, see the official [documentation of poetry](https://python-poetry.org/docs/cli/#version) for possible values |
+| python-version    |    ❌     |     3.10      |                  number                   | The python version for setting up poetry.                                                                                         |
+| poetry-version    |    ❌     |    1.1.13     |                  number                   | The poetry version to be installed.                                                                                               |
+| working-directory |    ❌     |       .       | The root directory of the poetry project. |                                                                                                                                   |
 
 ## Output Parameters
 
-| Name        | Description                                         |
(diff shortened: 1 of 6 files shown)


actions/python-setup-poetry/action.yaml
```diff
@@ -18,35 +18,28 @@ inputs:
...
-        python-version: ${{ inputs.python-version }}
-
-    - name: Get full Python version
-      id: full-python-version
-      run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
-      shell: bash
-
-    - name: Bootstrap poetry
+    - name: Install Poetry ${{ inputs.poetry-version }}
+      run: pipx install poetry==${{ inputs.poetry-version }}
```
yannick-roeder, line 22: Where does `pipx` come from and does it not depend on python being installed before? Are you using a specific docker image when using the action?

actions/python-poetry-bump-version/README.md
```diff
@@ -7,37 +7,18 @@ in the `pyproject.toml` file. It will output the old and new bumped versions.
...
-    with:
-      release-type: ${{ inputs.release-type }}
-      python-version: ${{ inputs.python-version }}
-      poetry-version: ${{ inputs.poetry-version }}
-
-  - name: Use bump version output
-    run: echo Bumped Version from ${{ steps.bump-version.outputs.old-tag }} to ${{ steps.bump-version.outputs.release-tag }}
-```
+| Name | Description |
+| ---- | ----------- |
```
yannick-roeder, line 24: Why did you delete the output parameters and sample usage section?

actions/python-setup-poetry/action.yaml
yannick-roeder: Interesting, so then the `setup-python` action below only configures python for venv with a specific version, but does not really install a system wide version like you would do locally

yannick-roeder: Can you find a more fitting description for this action in line 2 then please?

disrupted: yep, it only takes care of our project's virtualenv and that's a good thing imo. Python is already installed by default.

actions/python-poetry-bump-version/README.md
disrupted: must've been by accident, good thing you noticed. thanks!

actions/python-poetry-bump-version/action.yaml
```diff
@@ -30,7 +30,7 @@ runs:
   using: "composite"
   steps:
     - name: Set up Poetry version ${{ inputs.poetry-version }}
-      uses: bakdata/ci-templates/actions/python-setup-poetry@v1.0.0
+      uses: bakdata/ci-templates/actions/python-setup-poetry@v1.3.0
```
raminqaf, line 33: Don't forget to create this tag ;)

actions/python-setup-poetry/action.yaml
```diff
@@ -18,35 +18,28 @@ inputs:
...
-    - name: Ensure cache is healthy
-      if: steps.cache.outputs.cache-hit == 'true'
-      run: timeout 10s python -m poetry run pip --version || rm -rf ${{ inputs.working-directory }}/.venv
+    - run: poetry env use ${{ inputs.python-version }}
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+
+    - name: Get full Python version
+      id: full-python-version
+      run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
```
raminqaf, line 44: What is the purpose of this step? To output the python version? Do we still need it?
I noticed that you are not using this output anywhere. It is also not properly defined on top of the `action.yaml` file like this:
```yaml
outputs:
  version:
    description: "The full python version."
    value: ${{ steps.full-python-version.outputs.version }}
```
I would also recommend changing the name to something more concrete like `full-python-version`. Please also add it in the README.

disrupted: Wasn't sure either so I left it there. I am happy to remove it though, if it's not needed.

actions/python-poetry-bump-version/action.yaml
disrupted: yep, will do after it's merged 🙂 hopefully we'll have a proper release pipeline for this soon

PR #32, opened 2022-10-05
Support custom working directory for bumpversion action

disrupted:
tested in the playground

### DIFF
--- a/actions/bump-version/README.md
+++ b/actions/bump-version/README.md
@@ -9,8 +9,6 @@ You need a `.bumpversion.cfg` file in `working-directory` (repository root by de
 ```cfg
 [bumpversion]
 current_version = 0.0.1
-commit = False
-tag = False
 ```
 
 > **Note**
--- a/actions/bump-version/README.md
+++ b/actions/bump-version/README.md
@@ -4,7 +4,7 @@ This action will bump the version of your project according to the `.bumpversion
 
 ## Prerequisites
 
-You need a `.bumpversion.cfg` file in the root directory of your repository. A minimal configuration could look like this:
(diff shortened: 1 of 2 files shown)


actions/bump-version/action.yaml
```diff
@@ -5,6 +5,10 @@ inputs:
   release-type:
     description: "The type of the release (major, minor or patch)."
     required: true
+  working-directory:
+    description: "The root directory of the poetry project."
```
yannick-roeder, line 9: I don't think this necessarily refers to a root of a poetry project. Just use a similar description like you did in the Readme

actions/bump-version/README.md
```diff
@@ -1,56 +1,51 @@
...
+
+You need a `.bumpversion.cfg` file in `working-directory` (repository root by default). A minimal configuration could look like this:
+
 ```cfg
 [bumpversion]
 current_version = 0.0.1
-commit = True
-tag = True
+commit = False
+tag = False
```
yannick-roeder, line 13: Why did you change this?

disrupted: I thought it would be less confusing for the example in the readme, since this action doesn't do any of these things. I also added a disclaimer underneath to link to the `commit-and-push` action. Of course it can still be set to true. Are these two settings actually required? Otherwise I would just remove them altogether.

yannick-roeder: I'm not sure if they are required, but I think removing them is a good idea if you test that it still works then

disrupted: Can confirm, it works without those options set.

yannick-roeder (approved):

PR #35, opened 2022-10-13
Add Helm package action & Helm release workflow for GH pages

disrupted:


.github/workflows/helm-release.yaml
```diff
@@ -72,29 +73,46 @@ jobs:
...
-        uses: bakdata/ci-templates/actions/bump-version@v1.0.0
+        uses: bakdata/ci-templates/actions/bump-version@v1.5.2
         with:
           release-type: ${{ inputs.release-type }}
+          working-directory: ${{ inputs.charts-dir }}
+
+      - name: Helm package
+        uses: bakdata/ci-templates/actions/helm-package@v1.6.0
+        with:
+          charts-dir: ${{ inputs.charts-dir }}
```
torbsto, line 86: With the changes suggested above in mind, this could be the temporary directory for the gh-pages artifact (or an empty directory if there is no artifact yet).
This would also mean that you no longer include the source of the chart (`Chart.yaml`, `templates/`, `bumpversion.cfg`...).

.github/workflows/README.md
```diff
@@ -7,22 +7,18 @@ The following workflows can be found here:
...
-This workflow will lint all charts, bump the project version according to the `.bumpversion.cfg` file, create releases
-for all changed charts and provide an `index.yaml` for all packaged charts as a GitHub web page.
+This workflow will lint all charts, bump the project version according to the `.bumpversion.cfg` file, create an artifact and provide an updated `index.yaml` for all packaged charts on GitHub pages.
 
 ### Prerequisites
 
-Your helm charts need to be located inside the `charts` folder of your repository to use this workflow and you need
-a `.bumpversion.cfg` file in your root directory. Make sure to set the correct path for the `Chart.yaml` file. A minimal
-configuration could look like this:
+Your helm charts and `.bumpversion.cfg` need to be located inside the `charts-dir` folder of your repository (repository root by default) to use this workflow. A minimal
```
torbsto, line 14:
```suggestion
Your Helm charts and `.bumpversion.cfg` need to be located inside the `charts-dir` folder of your repository (repository root by default) to use this workflow. A minimal
```

.github/workflows/helm-release.yaml
```diff
@@ -17,16 +17,17 @@ on:
         required: false
         default: ".github/lint-config.yaml"
         type: string
-      python-version:
```
torbsto, line 20: You could still forward it to `bakdata/ci-templates/actions/bump-version@v1.5`

torbsto (commented):
Thanks! Looking very much forward to properly reference bakdata Helm charts :tada:

.github/workflows/helm-release.yaml
disrupted: our bump-version action no longer supports passing a specific Python version in the latest release

actions/helm-package/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+  steps:
+    - name: Set up Helm
+      uses: azure/setup-helm@v1
+      with:
+        version: "${{ inputs.helm-version }}"
+
+    - name: Helm package
+      run: |
+        helm package .
+        helm repo index .
```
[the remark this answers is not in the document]
disrupted: great suggestions! I will see if I can implement this logic in the workflow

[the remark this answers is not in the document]
yannick-roeder: Maybe, I'm missing something, but why do we need to unpack the current artifact after downloading from `gh-pages`? I thought we would have 1 artifact for every chart version on the `gh-pages` branch then, so only the `index.yaml` file gets updated. So `gh-pages` branch would look like this:
```
index.yaml
my-chart-0.0.1.tgz
my-chart-0.0.2.tgz
...
```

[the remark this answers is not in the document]
disrupted: @yannick-roeder Current pages artifact means the entire pages source bundled by the `upload-pages-artifact` action. This archive should contain all the files you mentioned.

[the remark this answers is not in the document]
yannick-roeder: Ah ok, that makes sense, thanks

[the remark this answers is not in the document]
disrupted: I looked into it more and downloading pages artifacts doesn't seem to be well supported.
Using the API one can at least list the existing artifacts
```
gh api repos/bakdata/<repo>/actions/artifacts --jq '[.artifacts[] | select(.name == "github-pages")] | first'
```
but they show as expired. Downloading is therefore not possible.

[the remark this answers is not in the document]
disrupted: example for the current artifact we would need to download

```jsonc
{
  ...
  "artifacts": [
    {
      "id": ...,
      "node_id": "...",
      "name": "github-pages",
      "size_in_bytes": 51200,
      "url": "https://api.github.com/repos/bakdata/<repo>/actions/artifacts/<artifact_id>",
      "archive_download_url": "https://api.github.com/repos/bakdata/<repo>/actions/artifacts/<artifact_id>/zip",
      "expired": true,
      "created_at": "2022-10-13T16:12:06Z",
      "updated_at": "2022-10-13T16:12:07Z",
      "expires_at": "2022-10-14T16:11:38Z",
      "workflow_run": {
        "id": ...,
        "repository_id": ...,
        "head_repository_id": ...,
        "head_branch": "main",
        "head_sha": "..."
      }
    },
  ]
}
```

```console
$ curl https://api.github.com/repos/bakdata/<repo>/actions/artifacts/<artifact_id>/zip
```

```json
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/actions#download-an-artifact"
}
```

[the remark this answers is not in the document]
disrupted: I have created a workaround:
1. download existing index.yaml from GH pages
2. parse all referenced package versions
3. Download old versions
4. Create new pages artifact including old versions

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
+
+      - name: Helm package
+        uses: bakdata/ci-templates/actions/helm-package@feat/helm-release
+        with:
+          charts-dir: ${{ inputs.charts-dir }}
 
-      - name: Commit and push bumpversion.cfg file
-        uses: bakdata/ci-templates/actions/commit-and-push@v1.0.0
+      - name: Commit and push Chart.yaml & .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@feat/helm-release
```
disrupted, line 98: change to release

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
           release-type: ${{ inputs.release-type }}
+          working-directory: ${{ inputs.charts-dir }}
+
+      - name: Download previous Helm index & packages
+        run: curl -O ${{ inputs.page-url }}/index.yaml
+        shell: bash
+        working-directory: ${{ inputs.charts-dir }}
+
+      - name: Helm package
+        uses: bakdata/ci-templates/actions/helm-package@feat/helm-release
```
disrupted, line 93: change to release

disrupted, line 93:
```suggestion
        uses: bakdata/ci-templates/actions/helm-package@v1.6.0
```

.github/workflows/helm-release.yaml
disrupted, line 98:
```suggestion
        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
```

actions/commit-and-push/action.yaml
```diff
@@ -18,17 +18,29 @@ inputs:
...
-        git commit -m "${{ inputs.commit-message }}" -a
+      shell: bash
+
+    - name: Add untracked files
+      run: git add .
+      if: ${{ inputs.add-untracked == 'true' }}
+      shell: bash
+
+    - name: Commit files
+      run: git commit -m "${{ inputs.commit-message }}" -a
```
torbsto, line 43:
```suggestion
      run: git commit -m "${{ inputs.commit-message }}"
```

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
           commit-message: "Bump version ${{ steps.release-tag.outputs.old-tag }} → ${{ steps.release-tag.outputs.release-tag }}"
           github-username: ${{ secrets.github-username }}
           github-email: ${{ secrets.github-email }}
           github-token: ${{ secrets.github-token }}
 
-      - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.2.1
+      - name: Find previous package versions
+        id: packages
+        uses: mikefarah/yq@master
```
torbsto, line 108: Why do we need master here?

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
         with:
           ref: ${{ inputs.ref }}
           commit-message: "Bump version ${{ steps.release-tag.outputs.old-tag }} → ${{ steps.release-tag.outputs.release-tag }}"
           github-username: ${{ secrets.github-username }}
           github-email: ${{ secrets.github-email }}
           github-token: ${{ secrets.github-token }}
 
-      - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.2.1
+      - name: Find previous package versions
```
torbsto, line 106: Really a nitpick, but maybe you could swap this with `Prepare artifact`. That would bring it closer to `Download packages` where its output is used and might help to follow the logic.

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
+        id: packages
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq '.entries.[] | map(.urls.[]) | join(",")' ${{ inputs.charts-dir }}/index.yaml
+
+      - name: Prepare artifact
+        run: |
+          mkdir artifact
```
torbsto, line 114: Also not a blocker, but it might be useful to make this an input parameter, so that users can configure it in case there is already a directory called artifact

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +77,68 @@ jobs:
...
       - name: Bump version
         id: release-tag
-        uses: bakdata/ci-templates/actions/bump-version@v1.0.0
+        uses: bakdata/ci-templates/actions/bump-version@v1.5.2
         with:
           release-type: ${{ inputs.release-type }}
+          working-directory: ${{ inputs.charts-dir }}
+
+      - name: Download previous Helm index & packages
+        run: curl -O ${{ inputs.page-url }}/index.yaml
```
torbsto, line 88: Did you check what happens if this doesn't exist?

torbsto: Okay, after a quick local check: This would create a file with the following content
```html
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

and therefore would lead to an error in the next step, because it's not JSON.

.github/workflows/helm-release.yaml
disrupted: I tested the latest tagged release (v4) first, but it gave an error. Only master seems to work.

.github/workflows/helm-release.yaml
disrupted: good point, I'll change it to skip if the file doesn't exist yet.

actions/commit-and-push/action.yaml
disrupted: I wouldn't like to change this as it's been the default before.
This makes sure that already tracked files are included in the commit. To include all changed files (untracked ones) I added a separate setting.

.github/workflows/helm-release.yaml
disrupted: Went for a more robust approach: If curl doesn't find a file, the step fails. For the initial setup of the workflow there is a `skip-download` input option, because there are no files yet on GH pages.

.github/workflows/helm-release.yaml
disrupted: this actually needs to be called earlier, before index.yaml is rebuilt. Otherwise, it will try to download the newly created package version from GH pages and fail.

actions/helm-package/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+  steps:
+    - name: Set up Helm
+      uses: azure/setup-helm@v3
+      with:
+        version: "${{ inputs.helm-version }}"
+
+    - name: Helm package
+      run: |
+        helm package .
+        helm repo index . --merge index.yaml
```
torbsto, line 24: In [Quick's case](https://github.com/bakdata/quick/blob/9063577615f8349e2ef47f1e1ebd52660640e327/.github/workflows/update-helm-chart.yml#L72), the index.yaml is in a different directory. So maybe:
```suggestion
        helm repo index . --merge ${{ inputs.index-location }}
```

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +87,72 @@ jobs:
...
         id: release-tag
-        uses: bakdata/ci-templates/actions/bump-version@v1.0.0
+        uses: bakdata/ci-templates/actions/bump-version@v1.5.2
         with:
           release-type: ${{ inputs.release-type }}
+          working-directory: ${{ inputs.charts-dir }}
 
-      - name: Commit and push bumpversion.cfg file
-        uses: bakdata/ci-templates/actions/commit-and-push@v1.0.0
+      - name: Download previous Helm index
```
torbsto, line 97: This and the following steps cover a lot of subtleties of gh-pages and Helm. I think a short overview explaining why all this is necessary would be helpful.

.github/workflows/README.md
```diff
@@ -7,22 +7,18 @@ The following workflows can be found here:
 
 ## Helm Release
```
torbsto, line 8: Not related to this PR, but the above listing of workflows in this repository is incomplete.

actions/helm-package/action.yaml
torbsto: On the other hand, one could also just move the index.yaml into the chart-directory for this action

.github/workflows/README.md
yannick-roeder: True, we forgot to continue the list. @disrupted can you complete the list in this PR please?

.github/workflows/helm-release.yaml
```diff
@@ -72,29 +87,72 @@ jobs:
...
+      - name: Prepare artifact
+        run: |
+          mkdir ${{ inputs.artifact-dir }}
+          cp index.yaml ${{ inputs.artifact-dir }}/
+          find . -maxdepth 1 -type f -name "*.tgz" -exec mv {} ${{ inputs.artifact-dir }} \;
+        shell: bash
+        working-directory: ${{ inputs.charts-dir }}
+
+      - name: Download packages
+        run: curl -O -f "${{ inputs.page-url }}/{${{ steps.packages.outputs.result }}}"
```
yannick-roeder, line 133: What is the reason for the third pair of curly braces here?
```suggestion
        run: curl -O -f "${{ inputs.page-url }}/${{ steps.packages.outputs.result }}"
```

.github/workflows/README.md
```diff
@@ -7,22 +7,18 @@ The following workflows can be found here:
...
+configuration with `charts-dir=charts` could look like this:
 
 ```cfg
 [bumpversion]
 current_version = 0.0.1
-commit = True
-tag = False
 
-[bumpversion:file:charts/my-chart/Chart.yaml]
+[bumpversion:file:charts/Chart.yaml]
```
yannick-roeder, line 21: The description above indicates that the workflow will work for multiple charts, but with this change only a single chart is supported in the `charts` directory. How would it work for multiple?

.github/workflows/README.md
```diff
@@ -106,11 +99,14 @@ jobs:
   call-workflow-passing-data:
     uses: bakdata/ci-templates/.github/workflows/helm-release.yaml@main
     with:
-      release-type: "${{ github.event.inputs.release-type }}"
+      release-type: ${{ inputs.release-type }}
       ref: "my-awesome-ref" # (Optional)
       lint-config-path: "my-lint-config.yaml" # (Optional)
-      helm-version: "v3.4.0" # (Optional)
+      helm-version: "v3.10.1" # (Optional)
       python-version: "3.8" # (Optional)
```
yannick-roeder, line 106: We do not support a python version anymore right?

.github/workflows/README.md
```diff
@@ -106,11 +99,14 @@ jobs:
   call-workflow-passing-data:
     uses: bakdata/ci-templates/.github/workflows/helm-release.yaml@main
     with:
```
yannick-roeder, line 101: What about the `page-url`?

actions/helm-lint/README.md
```diff
@@ -1,32 +1,34 @@
...
+| ---------------- | :------: | :-----------------------------------: | :----: | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| ref              |    ❌    | The default branch of your repository | string | The ref name to checkout the repository                                                                                                  |
+| lint-config-path |    ❌    |      ".github/lint-config.yaml"       | string | The path to the lint configuration file (For an example see https://github.com/helm/chart-testing/blob/main/pkg/config/test_config.yaml) |
+| helm-version     |    ❌    |               "v3.10.1"               | string | The Helm version                                                                                                                         |
 
 ## Usage
 
 ```yaml
-...
+---
```
yannick-roeder, line 26: Please remove. `---` is a separator for different yaml resources and would not work here

.github/workflows/README.md
disrupted: sorry, it doesn't work for multiple. Let me update the docs once more

.github/workflows/README.md
disrupted: indeed, I will remove it

.github/workflows/README.md
```diff
@@ -35,41 +31,38 @@ A minimal configuration could look like this:
...
-| python-version   |    ❌    |                "3.10"                 | string | The python version for bump2version                                                                                                        |
-| helm-version     |    ❌    |               "v3.4.0"                | string | The helm version                                                                                                                           |
+| helm-version     |    ❌    |               "v3.10.1"               | string | The Helm version                                                                                                                           |
+| charts-dir       |    ❌    |                  "."                  | string | The directory containing the Helm charts and `.bumpversion.cfg` file                                                                       |
+| skip-download    |    ❌    |                "false"                | string | Skip downloading index.yaml and previous Chart versions from GitHub pages. (To be used during setup of this workflow)                      |
```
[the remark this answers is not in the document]
disrupted: I think so. Unless you can do it from another branch. I remember GitHub didn't support calling a newly created workflow unless it exists on the main branch

actions/helm-package/action.yaml
disrupted: I would prefer keeping the complexity low at this point. we can always extend it later on

.github/workflows/helm-gke-deploy.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup Environment
-        uses: bakdata/ci-templates/actions/helm-gke-setup@v1.5.0
+        uses: bakdata/ci-templates/actions/helm-gke-setup@feat/helm-release
```
disrupted, line 79:
```suggestion
        uses: bakdata/ci-templates/actions/helm-gke-setup@v1.6.0
```

.github/workflows/helm-gke-destroy.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup Environment
-        uses: bakdata/ci-templates/actions/helm-gke-setup@v1.5.0
+        uses: bakdata/ci-templates/actions/helm-gke-setup@feat/helm-release
```
disrupted, line 59:
```suggestion
        uses: bakdata/ci-templates/actions/helm-gke-setup@v1.6.0
```

.github/workflows/helm-release.yaml
disrupted: Good point. I have added a paragraph in the readme explaining why this workaround is necessary and what it does.

.github/workflows/README.md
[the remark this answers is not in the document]
yannick-roeder: Ok, I would prefer a solution where the action even works when it cannot find something to download, but you can also resolve that in an issue

.github/workflows/README.md
yannick-roeder: This should then also be resolved as an issue

yannick-roeder (dismissed):

.github/workflows/README.md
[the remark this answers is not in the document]
disrupted: Created issue #37

yannick-roeder (approved):

PR #36, opened 2022-10-17
Fix Poetry setup on Windows

### LINKED ISSUE #33: Setup poetry template does not work with different python versions
I am trying to set up poetry with a Python version matrix from 3.7 to 3.9. Although I am passing the `python-version` input,  the `python-setup-poetry` template seems to ignore the python version and set up the poetry with Python 3.9.13. Below you can find the YAML file for the action and the action logs:

- [ci.yaml file](https://github.com/bakdata/quick-cli/blob/9969a292ed4cd899d64e7a08d429c16a121981aa/.github/workflows/ci.yaml)
- [Action logs](https://github.com/bakdata/quick-cli/actions/runs/3232853325/jobs/5293961410)

I am using the v1.5.2 version of the ci-templates. The workaround would be to downgrade the template to version 1.2.0.

disrupted:
Fixes #33

added a test workflow to verify that the correct versions of Python and Poetry are set up. It automatically runs when changes are made to the python-setup-poetry action or to the test workflow itself.

### DIFF
--- a/actions/python-setup-poetry/action.yaml
+++ b/actions/python-setup-poetry/action.yaml
@@ -1,28 +1,28 @@
-name: "Setup poetry"
-description: "Set up poetry with virtualenv caching"
+name: "Setup Poetry"
+description: "Set up Poetry with virtualenv caching"
 
 inputs:
   python-version:
-    description: "The python version for setting up poetry."
+    description: "The Python version for setting up Poetry."
     required: false
     default: "3.10"
   poetry-version:
-    description: "The poetry version to be installed."
+    description: "The Poetry version to be installed."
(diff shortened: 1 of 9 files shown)


actions/python-setup-poetry/action.yaml
```diff
@@ -30,11 +30,12 @@ runs:
 
     - name: Set up Python ${{ inputs.python-version }}
       uses: actions/setup-python@v4
+      id: python
```
raminqaf, line 33: Maybe `set-up-python` would be a bit clearer

raminqaf (approved):

PR #42, opened 2022-11-03
Add post-renderer option to helm-gke-deploy action and clean up documentation

yannick-roeder:


.github/workflows/helm-gke-deploy.yaml
```diff
@@ -88,12 +97,14 @@ jobs:
           helm-version: ${{ inputs.helm-version }}
 
       - name: Deploy Helm Chart
-        uses: bakdata/ci-templates/actions/helm-gke-deploy@v1.5.0
```
yannick-roeder, line 91: TODO: Update to proper version

actions/helm-gke-deploy/action.yaml
```diff
@@ -42,18 +48,19 @@ runs:
...
-            ${{ inputs.release-name }} ${{ inputs.chart }}
+          echo "Deploying with single value file"          
+          command="${command} --values ${{ inputs.values-yaml }}"
         fi
+
+        if [ -n "${{ inputs.chart-version }}" ]; then command="${command} --version ${{ inputs.chart-version }}"; fi
+        if [ -n "${{ inputs.post-renderer }}" ]; then command="${command} --post-renderer ${{ inputs.post-renderer }}"; fi
+
+        echo "Deploying with command: ${command}"
+        eval "$command"
```
torbsto, line 66: As you know I'm not a bash expert, but I thought using `eval` should be avoided, see for example [Google's shell guide](https://google.github.io/styleguide/shellguide.html#eval). You could instead collect all parameters in an array and then execute `helm "${parameters[@]}"`.

yannick-roeder: I did it without an array now. What do you think?

disrupted (approved):

actions/helm-gke-deploy/action.yaml
yannick-roeder: I used an array now

torbsto (approved):

PR #74, opened 2023-01-31
Update Poetry to 1.3.2

disrupted:


philipp94831 (commented):
Why do we need to update the version? Isn't the version customizable when using the workflows/actions?

disrupted:
> Why do we need to update the version? Isn't the version customizable when using the workflows/actions?

I would prefer setting a new default, because Poetry has a new lock file format. Otherwise we have to overwrite Poetry version in a lot of places.

disrupted:
> Hm, I don't think this is worth a PR. Poetry version should be pinned in other Repos anyways and you also would need to bump the workflow version in those repos anyways

ok, you're right

PR #84, opened 2023-03-20
Generate changelog for releases

### LINKED ISSUE #8: Action to create a changelog
This is currently missing from the Poetry publish workflow, but would make for a useful addition.

_Originally posted by @disrupted in https://github.com/bakdata/ci-templates/issues/3#issuecomment-999616588_

MichaelKora:
Fixes #8


.github/workflows/README.md
```diff
@@ -959,3 +959,55 @@ jobs:
     secrets:
       google-credentials: ${{ secrets.GKE_SA_TF }}
 ```
+
+## Generate Changelog
+
+This workflow will allow for the automatic generation of a changelog. The changelog template needs to be configured using a `changelog-config.json`. Depending on whether there is other documentation besides the GitHub changelog or not, one or two `Changelog` files will be updated.
```
raminqaf, line 965: Also good to link the GitHub action that we are using

.github/workflows/README.md
```diff
@@ -959,3 +959,55 @@ jobs:
...
+### Prerequisites
+
+Create a `changelog-config.json` file containing the configurations of the changelogs.
+
+### Input Parameters
+
+| Name          | Required |           Default Value           |  Type  | Description                                                                                                        |
+| ------------- | :------: | :-------------------------------: | :----: | ------------------------------------------------------------------------------------------------------------------ |
+| old-tag       |    ✅    |                 -                 | string | Previous version                                                                                                   |
+| new-tag       |    ✅    |                 -                 | string | Newest version                                                                                                     |
```
raminqaf, line 976:
```suggestion
| new-tag       |    ✅    |                 -                 | string | New version                                                                                                     |
```

.github/workflows/README.md
```diff
@@ -959,3 +959,55 @@ jobs:
...
+
+```yaml
+name: Github changelog creator
+on:
+  workflow_dispatch:
+
+jobs:
+  call-workflow-passing-data:
+    name: Create changelog
+    uses: bakdata/ci-templates/.github/workflows/generate-changelog.yaml@feat/changelog
```
yannick-roeder, line 1002: Please change to `@main` in docs

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+          { echo "content<<EOF"; echo "$content"; echo "EOF"; } >> "$GITHUB_OUTPUT"
+        shell: bash
+
+  generate-changelog:
+    name: Generate Changelog
+    runs-on: ubuntu-22.04
+    needs: [read-config-file]
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
```
raminqaf, line 71:
```suggestion
        uses: actions/checkout@v3.4
```

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+        shell: bash
+
+      - name: Create release commit
+        run: |
+          # Set email and username for github-actions bot
+          git config --local user.email "${{ secrets.github-email }}"
+          git config --local user.name "${{ secrets.github-token }}"
+          # Add changes
+          git add CHANGELOG.md
+          git add docs/
```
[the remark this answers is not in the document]
yannick-roeder: Also conisider using the complete path to the `Changelog.md` file inside the docs here

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+          # required for bypassing status checks when pushing
+          token: "${{ secrets.github-token }}"
+
+      - name: Read changelog config
+        id: read-config
+        # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+        run: |
+          content=$(cat "${{ inputs.config }}")
+          { echo "content<<EOF"; echo "$content"; echo "EOF"; } >> "$GITHUB_OUTPUT"
+        shell: bash
```
yannick-roeder, line 63: This can be be part of `generate-changelog`, no need for a separate job just to read a file from the repository

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+    name: Read config file
+    runs-on: ubuntu-22.04
+    outputs:
+      content: ${{ steps.read-config.outputs.content }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          # required for getting all tags
```
torbsto, line 52: You don't need all tags though, do you?

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+  generate-changelog:
+    name: Generate Changelog
+    runs-on: ubuntu-22.04
+    needs: [read-config-file]
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          # required for getting all tags
```
torbsto, line 74: see above

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+          fetch-depth: 0
+          # required for bypassing status checks when pushing
+          token: "${{ secrets.github-token }}"
+
+      - name: Generate Changelog
+        id: generate-release-changelog
+        uses: heinrichreimer/github-changelog-generator-action@v2.2
```
torbsto, line 81: Why did you decide to use this action? (@yannick-roeder) Both the action itself and the project it uses are somewhat dead.

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+          unreleased: "true"
+          pullRequests: "false"
+          compareLink: "true"
+          issuesWoLabels: "false"
+          futureRelease: "${{ inputs.new-tag }}"
+          issuesLabel: "**Miscellaneous:**"
+          bugLabels: "type/bug"
+          enhancementLabels: "type/enhancement"
+          httpCache: "true"
+          verbose: "true"
```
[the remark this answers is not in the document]
torbsto: It is definitely required, a lot of parameters don't work outside of Quick's setup

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+            sed -i '1d' changes.md
+            sed -i '/# Changelog/ r changes.md' "${{ inputs.doc-changelog }}"
+          fi
+        shell: bash
+
+      - name: Create release commit
+        run: |
+          # Set email and username for github-actions bot
+          git config --local user.email "${{ secrets.github-email }}"
+          git config --local user.name "${{ secrets.github-token }}"
```
torbsto, line 121: This leaks the token!

```suggestion
          git config --local user.name "${{ secrets.github-username }}"
```

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
...
+          # Add changes
+          git add CHANGELOG.md
+          git add docs/
+          git add deployment/
+          git commit -m "Release v${{ inputs.new-tag }}"
```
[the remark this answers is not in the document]
torbsto: I'd argue that this action should not create a commit at all (or at least make it optional). A lot of times you want to include other changes in the release commit.

torbsto (changes requested):

.github/workflows/generate-changelog.yaml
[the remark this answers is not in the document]
yannick-roeder: I was thinking about that as well, but so far we only have use cases where we commit and push it right away. How would you output it otherwise? Would you output it as a file not being committed or would you use an job outputs?

.github/workflows/generate-changelog.yaml
```diff
@@ -0,0 +1,127 @@
+name: Generate changelog
+# Reusable workflow to generate changelog after release
```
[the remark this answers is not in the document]
yannick-roeder: I think the name should be kept short and descriptive because it can appear in the GitHub UI and sadly a proper description field is not supported. So I prefer having a comment for the description

.github/workflows/generate-changelog.yaml
torbsto: No, it works fine for us, but maybe it makes sense to evaluate the current solutions w.r.t. the requirements of bakdata's different projects before committing.

.github/workflows/generate-changelog.yaml
MichaelKora: this CI pipeline is strongly based on the quick one. since i dont have any other use case to compare it with, i am using the same values quick used.

MichaelKora: i thought that is usually done that way when i dealing with changelog

torbsto: Okay, we use a fetch-depth of 0 there because we also have to compute the new tag. I'd assume you can omit it for this action.

.github/workflows/generate-changelog.yaml
raminqaf: I found this other action, which looks more alive to me https://github.com/mikepenz/release-changelog-builder-action/tree/v3.7.0

.github/workflows/README.md
```diff
@@ -959,3 +959,55 @@ jobs:
...
       google-credentials: ${{ secrets.GKE_SA_TF }}
 ```
+
+## Generate Changelog
+
+This workflow will allow for the automatic generation of a changelog. The changelog template needs to be configured using a `changelog-config.json`. Depending on whether there is other documentation besides the GitHub changelog or not, one or two `Changelog` files will be updated.
+
+### Prerequisites
+
+Create a `changelog-config.json` file containing the configurations of the changelogs.
```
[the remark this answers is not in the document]
MichaelKora: No, I haven't found anything reliable. I just looked at the config in the quick project you sent me.

.github/workflows/generate-changelog.yaml
MichaelKora: the current one offers more configurations and flexibility

.github/workflows/generate-changelog.yaml
yannick-roeder: Thanks @torbsto, I replaced the token with a new one. @MichaelKora please be careful where you set that token

.github/workflows/generate-changelog.yaml
yannick-roeder: It would be could if you could check how good the other 2 options would work for us

.github/workflows/generate-changelog.yaml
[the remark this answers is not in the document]
yannick-roeder: Please remove the git commit and push step. It is fine if you just leave the Changelog artifact(s) in the local file system

.github/workflows/README.md
```diff
@@ -928,6 +928,12 @@ This workflow leverages your Terraform project to provision GCP resources.
 
 Create a service account and save its key as a GitHub secret so you can provide it to the variable `google-credentials`.
 
+### Dependencies
+
+This workflow is built from multiple composite actions listed below:
+
+- [action-github-changelog-generator](https://github.com/heinrichreimer/action-github-changelog-generator)
```
raminqaf, line 935: Again I am a bit worried about this action.
https://github.com/heinrichreimer/action-github-changelog-generator/issues/42
Take the issue as an example, the issue for the depricated `set-ouput` command was opened but never fixed.

yannick-roeder: @MichaelKora is in the process of testing the other ones

MichaelKora:
am closing this resolving this conversion since is covered by the [previous](https://github.com/bakdata/ci-templates/pull/84#discussion_r1146085648) one.

.github/workflows/README.md
MichaelKora: am closing this resolving this conversion since is covered by the [previous](https://github.com/bakdata/ci-templates/pull/84#discussion_r1146085648) one.

.github/workflows/generate-changelog.yaml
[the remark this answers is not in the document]
MichaelKora: am resolving this conversation since this is implemented/fixed in the new version

actions/generate-changelog/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+  steps:
+    - name: "Create changelog"
+      id: build_changelog
+      uses: mikepenz/release-changelog-builder-action@v3.7.0
+      with:
+        token: "${{ inputs.github-token }}"
+        configuration: "${{ inputs.config }}"
+        fromTag: "${{ inputs.old-tag }}"
+        toTag: "${{ inputs.new-tag }}"
+        outputFile: "${{ inputs.output }}"
```
torbsto, line 49: It seems like this action has a number of output, among others `outputs.changelog`. Can we also use an output instead of writing it to a temporary file?

actions/generate-changelog/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+        fetchReviewers: "${{ inputs.fetchReviewers }}"
+        fetchReleaseInformation: "${{ inputs.fetchReleaseInformation }}"
+        commitMode: "${{ inputs.commitMode }}"
+
+    - name: Update github Changelog
```
torbsto, line 53: Do we need to check `outputs.failed`?

torbsto (commented):
Looks good!

actions/generate-changelog/action.yaml
MichaelKora: what do you mean?

actions/generate-changelog/action.yaml
MichaelKora: we could use it but the generated changelog is a kind of standalone changelog, because it just generate changelogs for the current tag, nbot for the previous..

actions/generate-changelog/action.yaml
MichaelKora: the `updste-step`  wont be executed if the  `created-one`. so i dont think we need to explicitly check that

actions/generate-changelog/action.yaml
torbsto: Yes, the temporary file was only a workaround. I think using the output is much cleaner because we have one less parameter and users don't need to worry that we add a file that they might accidentally commit and push

MichaelKora: if we do not want to modify the `Changelog` file in this composite action but want to return the raw `Changelog` instead, then we do not need a composite action at all. because the used action already does that.

MichaelKora:
I'm closing this PR since there's no persuasive reason to add any new features to the existing one.

actions/generate-changelog/action.yaml
torbsto: @MichaelKora @yannick-roeder  Just to clarify, I do want to update the changelog file, but I also need the generated changelog for the GitHub release later on, for example in Quick:
- Changelog update: https://github.com/bakdata/quick/blob/master/CHANGELOG.md#080-2022-10-14
- Release note: https://github.com/bakdata/quick/releases/tag/0.8.0

PR #95, opened 2023-04-26
Saving application version under _version.py file to support adding a…

szymonsadowski3:
Related to KPOPS pull request: https://github.com/bakdata/kpops/pull/200 which addreses https://github.com/bakdata/kpops/issues/199

The idea is to have a _version.py file containing the version which can be later on used to determine the version in the CLI command (see https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package)


disrupted:
closing as its no longer needed

PR #107, opened 2023-06-07
Allow the action to fetch all existing tags.

### LINKED ISSUE #104: Fix changelog generation by allowing to fetch older tags

MichaelKora:
Fixes #104

### DIFF
--- a/actions/changelog-generate/action.yaml
+++ b/actions/changelog-generate/action.yaml
@@ -47,7 +47,8 @@ runs:
     - name: "Set tags"
       id: setTags
       run: |
-        tag_list=$(git describe --tags) || tag_list=""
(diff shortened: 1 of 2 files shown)


torbsto (commented):
Could you also add documentation for the required setup of the checkout action?

actions/changelog-generate/action.yaml
```diff
@@ -47,7 +47,7 @@ runs:
     - name: "Set tags"
       id: setTags
```
disrupted, line 48: why are some of the variables (also action inputs) in camel case 👀 I would prefer following a consistent style

MichaelKora: you are right. the vars have been updated..

PR #110, opened 2023-06-14
Set bumpversion `--no-tag` option

disrupted:


sujuka99 (approved):
LGTM, but maybe wait for another approval, too

PR #116, opened 2023-06-20
Fix boolean evaluation for changelog creation

irux:
it seems it is not getting this as a string

https://github.com/bakdata/kpops/blob/main/.github/workflows/release.yaml#L23

https://github.com/actions/runner/issues/2238
https://github.com/actions/runner/issues/1483

it seems it only applies when the input is from the dispatch and not as a fixed param

### DIFF
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -101,6 +101,7 @@ jobs:
           github-username: ${{ secrets.github-username }}s
           github-email: ${{ secrets.github-email }}
           github-token: ${{ secrets.github-token }}
+          add-untracked: "true"
 
       - name: Tag and release
         uses: bakdata/ci-templates/actions/tag-and-release@v1.22.0
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -94,7 +94,7 @@ jobs:
           fetch-release-information: "true"
 
       - name: Commit and push pyproject.toml file
-        uses: bakdata/ci-templates/actions/commit-and-push@v1.25.0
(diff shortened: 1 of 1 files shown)


disrupted:
so this contradicts the two linked issues?

.github/workflows/python-poetry-release.yaml
```diff
@@ -94,13 +94,15 @@ jobs:
           fetch-release-information: "true"
 
       - name: Commit and push pyproject.toml file
-        uses: bakdata/ci-templates/actions/commit-and-push@v1.0.0
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.25.0
```
disrupted, line 97: we removed the leading `v` in tags, remember? 😉 
```suggestion
        uses: bakdata/ci-templates/actions/commit-and-push@1.25.0
```

.github/workflows/python-poetry-release.yaml
```diff
@@ -94,13 +94,15 @@ jobs:
...
         with:
           ref: ${{ inputs.ref }}
           commit-message: "Bump version ${{ steps.bump-version.outputs.old-version }} → ${{ steps.bump-version.outputs.release-version }}"
           github-username: ${{ secrets.github-username }}s
           github-email: ${{ secrets.github-email }}
           github-token: ${{ secrets.github-token }}
+          add-untracked: "true"
+
```
disrupted, line 105:
```suggestion
```

disrupted (changes requested):

.github/workflows/python-poetry-release.yaml
irux: not for that version xD

disrupted: oh right! 😄

disrupted (approved):

PR #121, opened 2023-07-04
Add action for recursively substituting environment variables

yannick-roeder:


actions/substitute-envs/action.yaml
```diff
@@ -0,0 +1,35 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Substitute envs
+      id: substitute
+      shell: bash
+      run: |
+        function substitute() {
+          # create (sub-)directories
+          mkdir -p "${{ runner.temp }}/$(dirname "$1")"
```
disrupted, line 21: consider using `mktemp -d`

disrupted: not sure there is an actual advantage in this case. it's usually safer to use `mktemp` than `mkdir` in scripts. feel free to dismiss if you think there aren't any concerns with using `mkdir`

disrupted (approved):

PR #123, opened 2023-07-12
Fix image name in build jib action

yannick-roeder:


actions/java-gradle-build-jib/action.yaml
```diff
@@ -41,7 +45,7 @@ runs:
         gradle-cache: ${{ inputs.gradle-cache }}
 
     - name: Build Docker image
-      run: ./gradlew --info --stacktrace jibBuildTar --image=${{ github.event.repository.name }}:${{ github.run_id }} -x test
+      run: ./gradlew --info --stacktrace jibBuildTar --image=${{ inputs.image-name }} -x test
```
philipp94831, line 48: Where did the run id go?

philipp94831: NVM, saw your comment

philipp94831 (approved):

actions/java-gradle-build-jib/action.yaml
philipp94831: @yannick-roeder But is it a problem that actions might assume a different local name (that one tih the run id)?

yannick-roeder: It is not ideal to set it when building since the image is uploaded as a GitHub artifact anyway. This way the tagging only takes place when publishing and not when building. We also did a similar change in this MR for the docker-build action: https://github.com/bakdata/ci-templates/commit/dd83b6efb1e7246a78b57fc1d3883ff09a2b32c7#diff-92a24487fb002d6c76b4c99c206e2ec2f111b8f55748349617c574e296d84b20L42-L44

philipp94831: Ok

yannick-roeder: The local folder shouldn't matter, because it is specified when the artifact is being downloaded again: https://github.com/bakdata/ci-templates/blob/main/actions/docker-publish/action.yaml#L36

And the `*.tar` file name shouldn't matter, because the publish action just expects any tar file: https://github.com/bakdata/ci-templates/blob/main/actions/docker-publish/action.yaml#L58

PR #124, opened 2023-07-24
Fix version grep in Java Gradle release

yannick-roeder:


disrupted:
don't we need to pre-bump the version in https://github.com/bakdata/ci-templates/blob/35db5578670d924b2fdbcf82a0f630c8cdaf6816/.github/workflows/java-gradle-release.yaml for the next release to include this fix?

actions/java-gradle-release/action.yaml
```diff
@@ -64,7 +64,7 @@ runs:
     - name: Bump version
       id: evaluate-version
       run: |
-        old_version=$(grep "version=" gradle.properties | cut -d "=" -f2)
+        old_version=$(grep "^version=" gradle.properties | cut -d "=" -f2)
```
philipp94831, line 67: Why don't you use `project_version=$(./gradlew properties -q | grep "^version:" | awk '{print $2}' | tr -d '[:space:]' | xargs)`. This let's you define the version in any way and it does not necessarily need to be in gradle.properties

yannick-roeder: It just makes the CI run much longer since Gradle is a requirement for the job that runs the action then, so it needs to be installed. Additional the project then gets build so this takes additional CI time. The building time can be reduced with proper caching, but it seems overkill just to grep a version

yannick-roeder: Wait, Gradle is installed anyway below for the release, so I'll change it

philipp94831 (dismissed):

philipp94831 (approved):

disrupted (approved):

PR #130, opened 2023-08-22
Allow use of lfs and caching

### LINKED ISSUE #126: Cache LFS files in usages of `actions/checkout`
GitHub has a minimal free quota for LFS https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage

It seems that `actions/checkout` does not cache LFS files between workflow runs in a repo when the action is used with `lfs: true` so that the quota is affected by every run: https://github.com/actions/checkout/issues/165

Consider using https://github.com/nschloe/action-cached-lfs-checkout which is able to cache LFS files.

MichaelKora:
Fixes #126


.github/workflows/checkout_repo.yaml
yannick-roeder, line 1: Let's just call it `checkout` like the original one

.github/workflows/checkout_repo.yaml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  workflow_call:
+    inputs:
+      lfs:
+        description: "Describes if the repository is using any LFS files"
+        required: false
+        default: "false"
+        type: string
+      cache:
+        description: "Describes if the repository is using any LFS files"
```
yannick-roeder, line 12:
```suggestion
      lfs-cache:
        description: "Whether to cache the LFS files in this repository. This variable has an impact only if lfs is set to true."
```

actions/checkout-repo/README.md
```diff
@@ -0,0 +1,40 @@
...
+This workflow is built from other composite actions listed below:
+
+- [actions/checkout](https://github.com/actions/checkout)
+- [actions/cache](https://github.com/actions/cache)
+
+### Input Parameters
+
+| Name  | Required | Default Value |  Type  | Description                                                                                                  |
+| ----- | :------: | :-----------: | :----: | ------------------------------------------------------------------------------------------------------------ |
+| lfs   |    ❌    |    "false"    | string | Define wether to use LFS.                                                                                    |
```
yannick-roeder, line 20:
```suggestion
| lfs   |    ❌    |    "false"    | string | Define whether to use LFS files.                                                                                    |
```

.github/workflows/checkout_repo.yaml
```diff
@@ -0,0 +1,48 @@
+name: Reusable workflow for checking out a repository
+
+on:
+  workflow_call:
+    inputs:
+      lfs:
+        description: "Describes if the repository is using any LFS files"
```
yannick-roeder, line 7:
```suggestion
        description: "Define whether to checkout LFS files"
```

.github/workflows/checkout_repo.yaml
```diff
@@ -0,0 +1,48 @@
...
+        required: false
+        default: "true"
+        type: string
+    outputs:
+      lfs-cache-hit:
+        description: "A boolean value to indicate an exact match was found for the key."
+        value: ${{ jobs.checkout-repo.outputs.lfs-cache-hit }}
+
+jobs:
+  checkout-repo:
```
yannick-roeder, line 22: Same here, please call it simply `checkout` everywhere. It is a bit cleaner

yannick-roeder (changes requested):

jakob-ed:
Nice work! Once this is merged, we can use this action instead of the standard `actions/cache` in actions where we already need to be able to checkout LFS files:
- https://github.com/bakdata/ci-templates/blob/e55bc07633adf0722cf3004da35209fd2418e5b8/actions/java-gradle-build-jib/action.yaml#L46
- https://github.com/bakdata/ci-templates/blob/e55bc07633adf0722cf3004da35209fd2418e5b8/actions/java-gradle-test/action.yaml#L52

.github/workflows/bump-version-release.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@v3
+        uses: bakdata/ci-templates/actions/checkout-repo@v1.26.0
```
yannick-roeder, line 57: Did you test it? Your action is called `checkout` now instead of `checkout-repo`, right?

yannick-roeder (changes requested):

.github/workflows/bump-version-release.yaml
MichaelKora: its tested with `bakdata/ci-templates/actions/checkout@feat/checkout-repo`. so i believe it should work once the release is done after merge

actions/checkout/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+outputs:
+  lfs-cache-hit:
+    description: "A boolean value to indicate an exact match was found for the key"
+    value: ${{ steps.cache-lfs.outputs.cache-hit }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check out repository
+      uses: bakdata/ci-templates/actions/checkout@v1.26.0
```
disrupted, line 42: isn't this self referencing?

MichaelKora: good catch..i updated it and tested it in pg

.github/workflows/bump-version-release.yaml
yannick-roeder: @MichaelKora You reverted this change again

actions/checkout/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+outputs:
+  lfs-cache-hit:
+    description: "A boolean value to indicate an exact match was found for the key"
+    value: ${{ steps.cache-lfs.outputs.cache-hit }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check out repository
+      uses: actions/checkout@v3
```
jakob-ed, line 42: Please make sure to update to `v4` before merge because of [issues with v3](https://github.com/actions/checkout/issues/1448)

yannick-roeder (approved):

.github/workflows/_format.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     name: dprint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: bakdata/ci-templates/actions/checkout@v1.26.0
```
yannick-roeder, line 11: This is not the most up to date version and also versions are without a `v` now

yannick-roeder (changes requested):

yannick-roeder (approved):

PR #131, opened 2023-08-22
Fix GKE authentication

disrupted:
Related to #133


disrupted:
solution doesn't work

Issue #90, opened 2023-04-06
Remove semver bumping from release-tag-version workflow


yannick-roeder:
We should be able to use this configuration for bumpversion to properly work with a `-snapshot` suffix without the need for semver:
```
[bumpversion]
current_version = 0.2.2-snapshot
search = version: {current_version}
replace = version: {new_version}
parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<release>snapshot))?
serialize = 
	{major}.{minor}.{patch}-{release}
	{major}.{minor}.{patch}

[bumpversion:part:release]
first_value = snapshot
optional_value = release
values = 
	snapshot
	release
```

Then `bumpversion release` always removes the suffix and `bumpversion major`, `bumpversion minor` and `bumpversion patch` always add it. In order to do a major or minor release 2 bumpversion commands can be chained then:
```shell
bumpversion release
bumpversion major/bumpversion minor
```

PR #136, opened 2023-08-29
New Actions And Workflow for bumping version 

yordanovsstoyan:
Fixes https://github.com/bakdata/argocd-apps-helm-chart/issues/9
Fixes https://github.com/bakdata/ci-templates/issues/90

### DIFF
--- a/.github/workflows/bump-version-release-stoyan.yaml
+++ b/.github/workflows/bump-version-release-stoyan.yaml
@@ -1,0 +1,113 @@
+name: Bump Version Release
+# Reusable workflow for creating release tags using bumpversion
+
+on:
+  workflow_call:
+    inputs:
+      release-type:
+        description: "Scope of the release (major, minor or patch)."
+        required: true
+        type: string
+      changelog:
+        description: "Create changelog for release."
+        required: false
+        default: false
+        type: boolean
+      changelog-config:
+        description: "Changelog config path."
+        required: false
+        type: string
+      working-directory:
+        description: "Working directory containing `.bumpversion.cfg`. (Default is .)"
+        required: false
+        default: "."
+        type: string
+
+    secrets:
+      github-username:
+        description: "The GitHub username for committing the changes."
+        required: true
+      github-email:
+        description: "The GitHub email for committing the changes."
+        required: true
+      github-token:
+        description: "The GitHub token for committing the changes."
+        required: true
+
+    # Map the workflow outputs to job outputs
+    outputs:
+      release-version:
+        description: "The bumped version."
+        value: ${{ jobs.release.outputs.release-version }}
+      old-version:
+        description: "The old version in your `.bumpversion.cfg` file."
+        value: ${{ jobs.release.outputs.old-version }}
+
+jobs:
+  release:
+    runs-on: ubuntu-22.04
+
+    # Map the job outputs to step outputs
+    outputs:
+      release-version: ${{ steps.bump-version.outputs.release-version }}
+      old-version: ${{ steps.bump-version.outputs.old-version }}
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false # required for pushing to protected branch later
+
+      - name: Bump version
+        id: bump-version
+        uses: bakdata/ci-templates/actions/bump-version-stoyan@feat/new-bumpversion
+        with:
+          release-type: ${{ inputs.release-type }}
+          working-directory: ${{ inputs.working-directory }}
+
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version.outputs.old-version }} → ${{ steps.bump-version.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
+
+      - name: Create changelog
+        id: build-changelog
+        uses: bakdata/ci-templates/actions/changelog-generate@1.25.4
+        if: ${{ inputs.changelog == 'true' }}
+        with:
+          github-token: ${{ secrets.github-token }}
+          config: ${{ inputs.changelog-config }}
+          new-tag: ${{ steps.bump-version.outputs.release-version }}
+          fetch-reviewers: "true"
+          fetch-release-information: "true"
+
+      - name: Tag and release
+        uses: bakdata/ci-templates/actions/tag-and-release@v1.22.0
+        with:
+          tag: "${{ steps.bump-version.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
+          release-title: "${{ steps.bump-version.outputs.release-version }}"
+          release-body: "${{ steps.build-changelog.outputs.single-changelog }}"
+
+      - name: Bump version
+        id: bump-version-2
+        uses: bakdata/ci-templates/actions/patch-snapshot@feat/new-bumpversion
+        with:
+          working-directory: ${{ inputs.working-directory }}
+          test-version: ${{ steps.bump-version.outputs.release-version }}
+
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version.outputs.release-version  }} → ${{ steps.bump-version-2.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
--- a/actions/bump-version-stoyan/action.yaml
+++ b/actions/bump-version-stoyan/action.yaml
@@ -1,0 +1,48 @@
+name: "Bump version"
+description: "Bump version with python bump2version using .bumpversion.cfg"
+
+inputs:
+  release-type:
+    description: "The type of the release (major, minor or patch)."
+    required: true
+  working-directory:
+    description: "The directory containing the `.bumpversion.cfg` file."
+    required: false
+    default: "."
+  new-version:
+    description: ""
+    required: false
+    default: ""
+
+outputs:
+  release-version:
+    description: "The bumped version of your project."
+    value: ${{ steps.bump-version.outputs.new-version }}
+  old-version:
+    description: "The old version in your `.bumpversion.cfg` file."
+    value: ${{ steps.bump-version.outputs.old-version }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up bump2version
+      run: |
+        pipx install bump2version
+      shell: bash
+
+    - name: Bump version
+      id: bump-version
+      run: |
+        parameters=(--no-commit --allow-dirty --no-tag ${{ inputs.release-type }})
+        test=(${{ inputs.release-type }})
+        echo "old-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
+        if [[ $test == patch ]]; then
+          parameters+=(--new-version ${{ inputs.new-version }})
+          bump2version --allow-dirty release
+        else
+          bump2version --allow-dirty "${parameters[@]}"
+          bump2version --allow-dirty release
+        fi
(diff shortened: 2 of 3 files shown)


.github/workflows/bump-version-release.yaml
```diff
@@ -95,3 +95,19 @@ jobs:
...
+        uses: bakdata/ci-templates/actions/patch-snapshot@feat/new-bumpversion
+        with:
+          working-directory: ${{ inputs.working-directory }}
+          test-version: ${{ steps.bump-version.outputs.release-version }}
+
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version.outputs.old-version  }} → ${{ steps.bump-version-snapshot.outputs.release-version }}"
```
torbsto, line 110: The example you linked produced wrong results:
- Bump version 0.0.4-snapshot → 0.0.4 
- Bump version 0.0.4-snapshot → 0.0.5-snapshot

but it should be:
- Bump version 0.0.4-snapshot → 0.0.4 
- Bump version 0.0.4 → 0.0.5-snapshot

Also, you added the second commit in this PR. Do we actually need this? Is there a requirement somewhere that specifies the desired result?

actions/bump-version/action.yaml
```diff
@@ -33,12 +33,16 @@ runs:
     - name: Bump version
       id: bump-version
       run: |
-        parameters=(--no-commit --no-tag ${{ inputs.release-type }})
+        parameters=(--no-commit --allow-dirty --no-tag ${{ inputs.release-type }})
+        test=(${{ inputs.release-type }})
```
torbsto, line 37: Not needed anymore?

actions/bump-version/action.yaml
```diff
@@ -33,12 +33,16 @@ runs:
...
         echo "old-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
-        if [ -n "${{ inputs.new-version }}" ]; then
+        if [[ $test == patch ]]; then
           parameters+=(--new-version ${{ inputs.new-version }})
+          bump2version --allow-dirty release
+        else
+          bump2version --allow-dirty "${parameters[@]}"
+          bump2version --allow-dirty release
```
torbsto, line 44: Could you add a comment here describing what this does and why we need to differentiate between the two cases?

torbsto (changes requested):

.github/workflows/bump-version-release.yaml
yordanovsstoyan: I fixed the output. Regarding the second commit, I could not get it working properly without it

actions/bump-version/action.yaml
yordanovsstoyan: Which one is not needed ?

torbsto: The test variable

yordanovsstoyan: I removed it and tested in the [repo](https://github.com/bakdata/github-playground-3)

yannick-roeder (changes requested):

yannick-roeder:
Could you add example documentation for setting up a repository that uses this workflow (e.g. contents of `.bumpversion.cfg` file)?

.github/workflows/bump-version-release.yaml
yordanovsstoyan: I had 2 commits before

actions/bump-version/action.yaml
```diff
@@ -33,12 +33,18 @@ runs:
     - name: Bump version
       id: bump-version
       run: |
-        parameters=(--no-commit --no-tag ${{ inputs.release-type }})
+        parameters=(--no-commit --allow-dirty --no-tag ${{ inputs.release-type }})
         echo "old-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
-        if [ -n "${{ inputs.new-version }}" ]; then
```
yannick-roeder, line 38: The option to set a new version was nice before. Ideally we can keep it around and it can just be the highest priority case in the if conditions. If new-version is not set as an input, then it makes sense to use our new logic

actions/bump-version/action.yaml
```diff
@@ -33,12 +33,18 @@ runs:
...
+          bump2version --allow-dirty release
+        else
+          bump2version --allow-dirty "${parameters[@]}"
+          bump2version --allow-dirty release
         fi
-        bump2version "${parameters[@]}"
         echo "new-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
+      # We use the case for release-type "patch", because if we do not we get inproper results. For example : 0.0.1-snapshot -> 0.0.3
+      # If release-type == patch , we use "bump2version release". This command basically removes the suffix. For example :  0.0.1-snapshot -> 0.0.1
+      # In the other case we bump the version (major or minor), then we remove the snapshot suffix
```
yannick-roeder, line 48: These comments are nice, but they are not quite in the correct place. Could you at them to the fitting if cases that they explain?

actions/bump-version/action.yaml
```diff
@@ -33,12 +33,18 @@ runs:
     - name: Bump version
       id: bump-version
       run: |
-        parameters=(--no-commit --no-tag ${{ inputs.release-type }})
+        parameters=(--no-commit --allow-dirty --no-tag ${{ inputs.release-type }})
```
yannick-roeder, line 36: Since we basically want to use the first 3 parameters for all of our bumpversion CLI calls, what do you think about removing `${{ inputs.release-type }}` from the list and adding that separately, e.g.:
```suggestion
        parameters=(--no-commit --allow-dirty --no-tag)
```
And then for every call below:
```bash
bump2version "${parameters[@]}" major/minor/patch/release
```

.github/workflows/bump-version-release.yaml
```diff
@@ -61,11 +61,20 @@ jobs:
...
           working-directory: ${{ inputs.working-directory }}
 
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version.outputs.old-version }} → ${{ steps.bump-version.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
```
yannick-roeder, line 76: I'm not sure why this was moved up, but this way it looks to me like the Changelog is likely not part of the Tag if it gets created after the push, right? Could you test with a CHANGLOG.md file?

.github/workflows/bump-version-release.yaml
```diff
@@ -95,3 +95,19 @@ jobs:
...
+          test-version: ${{ steps.bump-version.outputs.release-version }}
+
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@v1.6.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version.outputs.release-version  }} → ${{ steps.bump-version-snapshot.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
```
yannick-roeder, line 113: We currently need 2x commit and push in this workflow, because we are creating the GitHub release in between, which is not ideal. As the [docs of the softprops/action-gh-release](https://github.com/softprops/action-gh-release?tab=readme-ov-file#-limit-releases-to-pushes-to-tags) action (which is part of our tag-and-release action above) mention, it would be more ideal if the GitHub release is actually created in the workflow running on the tag later with an `if: startsWith(github.ref, 'refs/tags/')` condition. This way we could avoid pushing and in turn triggering CI multiple times on the defualt branch.

actions/bump-version/action.yaml
```diff
@@ -25,20 +41,28 @@ outputs:
...
+        echo "old-version=$(bump-my-version show current_version | tail -1)" >> "$GITHUB_OUTPUT"
+
+        if [[ ${{ inputs.new-version }} != "" ]]; then
+          bump-my-version --new-version ${{ inputs.new-version }}
+        else
+          bump-my-version "${parameters[@]}" ${{ inputs.release-type }}
         fi
-        bump2version "${parameters[@]}"
-        echo "new-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
+        echo "bump-my-version show current_version | tail -1)" >> "$GITHUB_OUTPUT"
```
yannick-roeder, line 60: I think you are missing `$(`, right? Please test:
```suggestion
        echo "$(bump-my-version show current_version | tail -1)" >> "$GITHUB_OUTPUT"
```

actions/bump-version/action.yaml
yannick-roeder: Shouldn't this be used in basically all bump-my-version commands now?

yannick-roeder (changes requested):

DerTiedemann:
@yannick-roeder  regarding https://github.com/bakdata/ci-templates/pull/136#discussion_r1705191494 (cant comment for some reason)
I left them in to be specifc and `--allow-dirty` is still required

.github/workflows/bump-version-release.yaml
```diff
@@ -57,34 +52,35 @@ jobs:
         uses: bakdata/ci-templates/actions/checkout@1.32.0
         with:
           ref: ${{ github.event.repository.default_branch }}
-          persist-credentials: false # required for pushing to protected branch later
+          persist-credentials: true # required for pushing to protected branch later
```
yannick-roeder, line 55: Why did this change? I think this would lead to the CI being logged in with default credentials and not with the bot user

.github/workflows/bump-version-release.yaml
```diff
@@ -57,34 +52,35 @@ jobs:
...
-          release-type: ${{ inputs.release-type }}
+          release-type: "release"
           working-directory: ${{ inputs.working-directory }}
 
+      - name: Commit .bumpversion.cfg file without snapshot
+        run: |
+          git add .bumpversion.cfg
+          git commit -m "Bump version ${{ steps.bump-version.outputs.old-version }} → ${{ steps.bump-version.outputs.release-version }}"
+          git fetch origin
+          git rebase --strategy-option=theirs origin/main
```
yannick-roeder, line 74: Why is there a need for rebasing if we checked out right before?

.github/workflows/bump-version-release.yaml
```diff
@@ -57,34 +52,35 @@ jobs:
...
+          git push --verbose
+
       - name: Create changelog
         id: build-changelog
-        uses: bakdata/ci-templates/actions/changelog-generate@1.35.0
+        uses: bakdata/ci-templates/actions/changelog-generate@tiedemann/changelog-action
```
yannick-roeder, line 79: Did you test the case where we also want to have the up to date snapshot in a file on the default branch and the tag?

actions/bump-version/action.yaml
```diff
@@ -1,9 +1,25 @@
...
+# serialize =
+# 	{major}.{minor}.{patch}-{release}
+# 	{major}.{minor}.{patch}
+
+# [bumpversion:part:release]
+# first_value = snapshot
+# optional_value = release
+# values =
+# 	snapshot
+# 	release
```
yannick-roeder, line 18: Nice that there is a comment here, ideally you can also put this in the docs

actions/bump-version/action.yaml
```diff
@@ -25,20 +41,27 @@ outputs:
...
         fi
-        bump2version "${parameters[@]}"
-        echo "new-version=$(python -c "from configparser import ConfigParser; cfg = ConfigParser(); cfg.read('.bumpversion.cfg'); print(cfg['bumpversion']['current_version'])")" >> "$GITHUB_OUTPUT"
+        echo "new-version=$(bump-my-version show current_version | tail -1)" >> "$GITHUB_OUTPUT"
+      # release: a.b.c-snapshot -> a.b.c
+      # release a.b.c -> crash
+      # major: a.b.c(-snapshot)? -> a+1.0.0-snapshot
+      # minor: a.b.c(-snapshot)? -> a.b+1.0-snapshot
+      # patch: a.b.c(-snapshot)? -> a.b.c+1-snapshot
+      # TLDR: release removes the snapshot suffix
```
yannick-roeder, line 65: Same as above, nice comments, but please also add it in the docs

yannick-roeder (changes requested):

.github/workflows/bump-version-release.yaml
yannick-roeder: I still think we shouldn't create the GitHub release in the CI running on the default branch

Issue #25, opened 2022-08-30
Replace changelog plugin in Java actions with changelog action


yannick-roeder:
## Description
Currently we are using the [org.hildan.github.changelog Gradle plugin](https://plugins.gradle.org/plugin/org.hildan.github.changelog) to generate changelogs when releasing Java Gradle projects. To align the changelog to all of our other projects and remove the additional dependency in every Gradle project we should replace the changelog generation in the [java-gradle-release action](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release/action.yaml#L89-L95) and the [java-gradle-release-github action](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release-github/action.yaml#L51-L57) with our [changelog-generate action](https://github.com/bakdata/ci-templates/blob/main/actions/changelog-generate/action.yaml).

## Acceptance criteria
- [ ] The changelog in the [java-gradle-release action](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release/action.yaml#L89-L95) is generated with our [changelog-generate action](https://github.com/bakdata/ci-templates/blob/main/actions/changelog-generate/action.yaml)
- [ ] The changelog in the [java-gradle-release-github action](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release-github/action.yaml#L51-L57) is generated with our [changelog-generate action](https://github.com/bakdata/ci-templates/blob/main/actions/changelog-generate/action.yaml)
- [ ] The changelog should contain similar information to the one created so far in our Gradle projects (e.g. https://github.com/bakdata/streams-bootstrap/blob/master/CHANGELOG.md)

yannick-roeder:
Related to https://github.com/bakdata/ci-templates/issues/8

yannick-roeder:
Closed by https://github.com/bakdata/ci-templates/pull/146

PR #141, opened 2023-09-07
Update changelog template and introduce codeowners to notify when a PR is created

### LINKED ISSUE #140: Use template changelog config by default
The changelog generation should "just work" with the default https://github.com/bakdata/ci-templates/blob/940e8e7e4bc3368a6496b0d50af180e86f56eb4b/.github/changelog-config.json without needing to create this file in every repository. Use placeholders `#{{OWNER}}` and `#{{REPO}}` in the releases url. https://github.com/mikepenz/release-changelog-builder-action#template-placeholders.

MichaelKora:
Fixes #140


actions/changelog-generate/action.yaml
```diff
@@ -75,12 +75,23 @@ runs:
         echo  "fromtag=$setfrom">> $GITHUB_OUTPUT
       shell: bash
 
+    - name: Download default changelog
+      id: config
+      run: |
+        if [[ -n "${{ inputs.config }}" ]]; then
+          echo  "path=${{ inputs.config }}">> $GITHUB_OUTPUT
+        else
+          wget -c https://raw.githubusercontent.com/bakdata/ci-templates/blob/1.33.0/.github/changelog-config.json -O default-config.json
```
yannick-roeder, line 84: Tag branches don't contain `blob` inside their path, you can test this with https://raw.githubusercontent.com/bakdata/ci-templates/1.32.0/.github/changelog-config.json
```suggestion
          wget -c https://raw.githubusercontent.com/bakdata/ci-templates/1.33.0/.github/changelog-config.json -O default-config.json
```

yannick-roeder (changes requested):

actions/changelog-generate/action.yaml
```diff
@@ -75,12 +75,23 @@ runs:
         echo  "fromtag=$setfrom">> $GITHUB_OUTPUT
       shell: bash
 
+    - name: Download default changelog
+      id: config
+      run: |
+        if [[ -n "${{ inputs.config }}" ]]; then
+          echo  "path=${{ inputs.config }}">> $GITHUB_OUTPUT
```
yannick-roeder, line 82:
```suggestion
          echo "path=${{ inputs.config }}" >> "$GITHUB_OUTPUT"
```

actions/changelog-generate/action.yaml
```diff
@@ -75,12 +75,23 @@ runs:
...
+        else
+          wget -c https://raw.githubusercontent.com/bakdata/ci-templates/1.33.0/.github/changelog-config.json -O default-config.json
+          echo  "path=default-config.json">> $GITHUB_OUTPUT
```
yannick-roeder, line 85:
```suggestion
          echo "path=default-config.json" >> "$GITHUB_OUTPUT"
```

yannick-roeder (approved):

actions/changelog-generate/action.yaml
```diff
@@ -21,7 +21,7 @@ inputs:
   config:
     description: Path to the changelog config JSON file
     required: false
-    default: "./.github/changelog-config.json"
+    default: ""
```
jakob-ed, line 24: What about moving `.github/changelog-config.json` to `actions/changelog-generate/changelog-config.json` and adapting the default like in the suggestion? I think then you could get rid of the "Download default changelog" step completely and also don't need to worry about tags there.

```suggestion
    default: "$GITHUB_ACTION_PATH/changelog-config.json"
```

MichaelKora: @jakob-ed  @yannick-roeder after testing that i have to say this is definitely a good approach. we should consider this in other use cases. the current use case is different because we use some variable in the config and the config only scrap the data in the repo where it's stored. This causes the URL not be generated. and that is the reason i dont think this is a good match for now

.github/workflows/java-gradle-release.yaml
```diff
@@ -62,7 +68,7 @@ jobs:
     steps:
       - name: Release on Github
         id: release
-        uses: bakdata/ci-templates/actions/java-gradle-release@1.27.0
+        uses: bakdata/ci-templates/actions/java-gradle-release@feat/changelog-def
```
MichaelKora, line 71: TODO: update branch

actions/java-gradle-release/action.yaml
```diff
@@ -70,33 +80,28 @@ runs:
...
-    - name: Generate changelog
-      run: ./gradlew -Pchangelog.releaseVersion=${{ steps.evaluate-version.outputs.release-version }} --stacktrace --info generateChangelog
-      shell: bash
-      working-directory: ${{ inputs.working-directory }}
-      env:
-        GITHUB_USER: ${{ inputs.github-username }}
-        GITHUB_TOKEN: ${{ inputs.github-token }}
+    - name: Create changelog
+      id: build-changelog
+      uses: bakdata/ci-templates/actions/changelog-generate@feat/changelog-def
```
MichaelKora, line 91: TODO: update branch

actions/changelog-generate/action.yaml
yannick-roeder: So if I understand you correctly, if we would move the `changelog-config.json` file to `actions/changelog-generate/changelog-config.json`, then the template variables, e.g. `${{REPO}}` or `${{OWNER}}` are either empty or not being set correctly then?

jakob-ed: > we use some variable in the config and the config only scrap the data in the repo where it's stored
Are you sure about this? In 2 out of 3 occurrences of changelog-config.json in this PR (among them actions/changelog-generate/changelog-config.json), there is `https://github.com/bakdata/ci-templates/releases/tag/` hard-coded now. Could that be the reason no variable resolution is happening?

MichaelKora: correct @yannick-roeder

jakob-ed: I just noticed we have `${{OWNER}}`, `${{REPO}}` in the changelog-config.json files. It should be `#{{OWNER}}`, `#{{REPO}}` according to the docs: https://github.com/mikepenz/release-changelog-builder-action#template-placeholders 
I guess if you use `${{...}}`, Github CI tries to resolve the templating and so that the `release-changelog-builder-action` can resolve it, it should be `#{{...}}`

yannick-roeder: I was wondering about the discrepancy from the issue and the code as well already

yannick-roeder: Although `${{TO_TAG}}` and `${{TITLE}}` seemed to have worked so far without issues

jakob-ed: Ah they recently changed this for v4: https://github.com/mikepenz/release-changelog-builder-action/releases/tag/v4.0.0

MichaelKora: ```I was wondering about the discrepancy from the issue and the code as well already```
@yannick-roeder oh yes i agree. i added it here because we have another draft issue and wanted to included it to test it all

actions/changelog-generate/action.yaml
```diff
@@ -73,14 +74,17 @@ runs:
           fi
         fi
         echo  "fromtag=$setfrom">> $GITHUB_OUTPUT
+        echo "path=${{ inputs.config }}" >> "$GITHUB_OUTPUT"
       shell: bash
 
     - name: "Create changelog"
+      env:
+        CONFIG_PATH: ${{ inputs.config }}
```
jakob-ed, line 82: Appears to be unused?

actions/java-gradle-release/action.yaml
```diff
@@ -2,18 +2,31 @@ name: "Release Java artifacts"
 description: "Release Java Gradle artifacts on Github"
```
yannick-roeder, line 2: Is this supposed to be part of this PR? According to your description this is not related to https://github.com/bakdata/ci-templates/issues/25

actions/java-gradle-release/action.yaml
```diff
@@ -70,33 +75,27 @@ runs:
...
-
-    - name: Create release
-      run: ./gradlew release -x test -Prelease.useAutomaticVersion=true -Prelease.releaseVersion=${{ steps.evaluate-version.outputs.release-version }} -PdisablePushToRemote=true
-      shell: bash
-      working-directory: ${{ inputs.working-directory }}   
+      id: bump-version
+      uses: bakdata/ci-templates/actions/bump-version@1.32.0
+      with:
+        release-type: ${{ inputs.release-type }}
+        working-directory: ${{ inputs.working-directory }}
```
yannick-roeder, line 82: Did you test this? This will not work for Gradle projects as they use a Gradle plugin for versioning and the version number can be found in a `gradle.properties` file, e.g.: https://github.com/bakdata/streams-bootstrap/blob/master/gradle.properties#L1

actions/changelog-generate/action.yaml
```diff
@@ -73,14 +74,15 @@ runs:
           fi
         fi
         echo  "fromtag=$setfrom">> $GITHUB_OUTPUT
+        echo "path=${{ inputs.config }}" >> "$GITHUB_OUTPUT"
```
yannick-roeder, line 77: Why is the remapping to an output here needed instead of just `configuration: "${{ inputs.config }}"` below?

yannick-roeder (changes requested):

actions/changelog-generate/action.yaml
yannick-roeder: As discussed, it is fine like this, because without the remapping it does not seem to work. Please leave a short comment above it, so we are not confused in the future

actions/java-gradle-release/action.yaml
MichaelKora: oh i used a .bumpversion.cfg to test it. you're right it wont work. should i update the bump action by adding a new var for the config ?

yannick-roeder: No, the bumping needs to happen with Gradle and we don't want to change that

yannick-roeder: This is why Gradle still needs to be part of this action, but can be removed in the [java-gradle-release-github](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release-github/action.yaml#L51-L57) action when you remove the changelog there

MichaelKora: so you want to just change the [changelog step](https://github.com/bakdata/ci-templates/blob/main/actions/java-gradle-release/action.yaml#L89-L95)?

yannick-roeder: In this action yes

MichaelKora: ok makes sense

actions/java-gradle-release/action.yaml
MichaelKora: right..that is due to the fact that in order to consider the PR related to using default for changelog i needed to test it. I need a proper repo to test it. ANd while testign i reliased that i cannot just test here because the azure pipeline is used..that's how i ended modifying the graddle action as well

MichaelKora: i could either rename the PR or separate the graddle related changes

yannick-roeder (approved):

PR #142, opened 2023-09-11
Add timeout var and force fail when timeout is reached

### LINKED ISSUE #132: Port-Forward action does not fail when timeout is reached
Additionally timeout period is currently statically set to 120 seconds. Ideally we can create an input and set the default to 30 seconds.

## Acceptance criteria

- [ ] Port forward properly fails when timeout is reached
- [ ] Timeout can be configured via input
- [ ] Default timeout is set to 30 seconds

MichaelKora:
Fixes #132

### DIFF
--- a/actions/port-forward/action.yaml
+++ b/actions/port-forward/action.yaml
@@ -14,6 +14,10 @@ inputs:
     description: kubectl version, e.g. `v1.23.0` or `latest`
     default: latest
     required: false
+  timeout:
+    description: Timeout for portfowrd
+    default: "30"
+    required: false
 
 runs:
   using: composite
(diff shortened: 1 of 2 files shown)


actions/port-forward/action.yaml
```diff
@@ -1,18 +1,22 @@
...
+  service:
+    description: Name of Kubernetes service
     required: true
-  kubectl-version:
-    description: kubectl version, e.g. `v1.23.0` or `latest`
-    default: latest
+  timeout:
+    description: Timeout for portfowrd
+    default: "30"
     required: false
```
yannick-roeder, line 20: Why did you change the order, wasn't it logical before? At least required inputs should be first

actions/port-forward/action.yaml
```diff
@@ -24,15 +28,16 @@ runs:
         version: ${{ inputs.kubectl-version }}
 
     - name: Establish port forward
-      shell: bash
+      env:
+        TIMEOUT: "${{ inputs.timeout }}"
```
yannick-roeder, line 32: Is a remap required here?

actions/port-forward/action.yaml
```diff
@@ -24,15 +28,16 @@ runs:
...
       run: |
         kubectl port-forward "services/${{ inputs.service }}" "${{ inputs.port }}:${{ inputs.port }}" --namespace ${{ inputs.namespace }} --kubeconfig ~/.kube/config > port-forward.out &
         # wait up to 60 seconds for port forwarding
-        for _ in {1..120}; do if grep -q "Forwarding from 127.0.0.1:${{ inputs.port }} -> ${{ inputs.port }}" port-forward.out; then
+        for _ in {1..$TIMEOUT}; do if grep -q "Forwarding from 127.0.0.1:${{ inputs.port }} -> ${{ inputs.port }}" port-forward.out; then
           echo "Port forwarding done"
           break
         else
-          cat port-forward.out || true
```
yannick-roeder, line 35: Where did this line go? This was usesful for seeing errors from the port-forward

actions/port-forward/action.yaml
```diff
@@ -24,15 +28,16 @@ runs:
...
-          echo "Waiting for port forwarding..."
-          sleep 0.5
+          echo "Portforwarding failed: Timeout reached"
+          break
```
yannick-roeder, line 41: This breaks the loop, but it doesn't let the CI run fail, right? The issue states that the CI run should fail when port forwarding cannot be established

yannick-roeder (changes requested):

actions/port-forward/action.yaml
MichaelKora: it just makes it easier to use in the bash code.

yannick-roeder (approved):

PR #147, opened 2023-09-27
Enable automatic changelog generation in release workflows

disrupted:
Followup from #141


.github/workflows/bump-version-release.yaml
```diff
@@ -11,11 +11,12 @@ on:
       changelog:
         description: "Create changelog for release."
         required: false
-        default: false
+        default: true
         type: boolean
       changelog-config:
         description: "Changelog config path."
         required: false
+        default: "$GITHUB_ACTION_PATH/changelog-config.json"
```
MichaelKora, line 19: am not sure if that'd work like this. i think in order for this to work we need to add the config file in the `bump-version-release` action dir..

.github/workflows/python-poetry-release.yaml
```diff
@@ -25,11 +25,12 @@ on:
       changelog:
         description: "Create changelog for release."
         required: false
-        default: false
+        default: true
         type: boolean
       changelog-config:
         description: "Changelog config path."
         required: false
+        default: "$GITHUB_ACTION_PATH/changelog-config.json"
```
MichaelKora, line 33: same as above

.github/workflows/bump-version-release.yaml
disrupted: you're right. after reading the [docs on GitHub variables](https://docs.github.com/en/actions/learn-github-actions/variables) I believe `$GITHUB_ACTION_PATH` actually doesn't work at all outside of a composite action... :/

disrupted: I've implemented a workaround

actions/changelog-generate/action.yaml
```diff
@@ -21,7 +21,7 @@ inputs:
   config:
     description: Path to the changelog config JSON file
     required: false
-    default: "$GITHUB_ACTION_PATH/changelog-config.json"
+    default: "default"
```
yannick-roeder, line 24: Why not use an empty string as the default here? That would also solve the issue with having to set defaults in every workflow that uses the action

yannick-roeder (changes requested):

actions/changelog-generate/action.yaml
disrupted: good plan

disrupted: changed

yannick-roeder (approved):

PR #154, opened 2023-10-02
Fix bug in changelog generation

### LINKED ISSUE #148: Fix bug in changelog generation
### Description
Currently `merged-changelog<<EOF` appears inside the generated change log of a GitHub release, which shouldn't be the case. This also seems to cause weird formatting. It is unclear if this also happens when the changelog is being written into a file.

### Acceptance criteria
- [ ] The changelog in a GitHub release is generated without `merged-changelog<<EOF` in its content
- [ ] The changelog in a GitHub release is correctly formatted
- [ ] The changelog in a markdown file is generated without `merged-changelog<<EOF` in its content
- [ ] The changelog in a markdown file is correctly formatted

MichaelKora:
Fixes #148

### DIFF
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -105,7 +105,6 @@ jobs:
           add-untracked: "true"
       
       - name: Test changelog
-        if: ${{ steps.set-tags.outputs.totag != inputs.new-tag }}
         run: |
           echo "================="
           echo ${{ steps.build-changelog.outputs.single-changelog }}
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -103,13 +103,6 @@ jobs:
           github-email: ${{ secrets.github-email }}
           github-token: ${{ secrets.github-token }}
           add-untracked: "true"
-      
-      - name: Test changelog
-        run: |
-          echo "================="
-          echo ${{ steps.build-changelog.outputs.single-changelog }}
-          echo "================="
-        shell: bash
 
       - name: Tag and release
         uses: bakdata/ci-templates/actions/tag-and-release@v1.22.0
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -85,8 +85,7 @@ jobs:
 
       - name: Create changelog
         id: build-changelog
-        uses: bakdata/ci-templates/actions/changelog-generate@feat/changelog-eof
-        # uses: bakdata/ci-templates/actions/changelog-generate@1.36.0
+        uses: bakdata/ci-templates/actions/changelog-generate@1.37.0
         if: ${{ inputs.changelog == true}}
         with:
           github-token: ${{ secrets.github-token }}
--- a/.github/workflows/python-poetry-release.yaml
+++ b/.github/workflows/python-poetry-release.yaml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Create changelog
         id: build-changelog
-        uses: bakdata/ci-templates/actions/changelog-generate@1.35.0
+        uses: bakdata/ci-templates/actions/changelog-generate@feat/changelog-eof
         if: ${{ inputs.changelog == true}}
         with:
           github-token: ${{ secrets.github-token }}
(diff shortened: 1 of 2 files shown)


actions/changelog-generate/action.yaml
```diff
@@ -74,16 +73,11 @@ runs:
           fi
         fi
         echo "fromtag=$setfrom">> $GITHUB_OUTPUT
-        if [ -n "${{ inputs.config }}" ]; then
+        if [  "${{ inputs.config }}" != '']; then
```
disrupted, line 76: can't you check for non-empty using `-z`?

disrupted (changes requested):

.github/workflows/python-poetry-release.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
 
       - name: Create changelog
         id: build-changelog
-        uses: bakdata/ci-templates/actions/changelog-generate@1.35.0
+        uses: bakdata/ci-templates/actions/changelog-generate@feat/changelog-eof
```
yannick-roeder, line 88: Todo

actions/changelog-generate/action.yaml
```diff
@@ -74,16 +73,14 @@ runs:
           fi
         fi
         echo "fromtag=$setfrom">> $GITHUB_OUTPUT
-        if [ -n "${{ inputs.config }}" ]; then
-          path="$GITHUB_ACTION_PATH/changelog-config.json)"
+        # if [  "${{ inputs.config }}" != '']; then
+        if [  -z "${{ inputs.config }}" ]; then
```
yannick-roeder, line 76:
```suggestion
        if [ -z "${{ inputs.config }}" ]; then
```

yannick-roeder (changes requested):

.github/workflows/python-poetry-release.yaml
MichaelKora: thanks... 
will do it if you think everything else is ok.

actions/changelog-generate/action.yaml
```diff
@@ -74,16 +73,11 @@ runs:
...
+        if [ -z "${{ inputs.config }}" ]; then
+          path="$GITHUB_ACTION_PATH/changelog-config.json"
         else
           path=${{ inputs.config }}
         fi
-        if [ ! -f "$path" ]; then
-          echo "changelog config file $path doesn't exist"
-          exit 1
-        fi
-        echo "path=$path" >> "$GITHUB_OUTPUT"
```
disrupted, line 86: why did you remove this check and how does the `path` output variable work now?

MichaelKora: the check was removed because it was not working as expected. the path works basically the same.. if a config was provided then it will be used; else the one from the action will be used

disrupted: what if a config path was provided but there's a mistake and the file doesn't actually exist?
also I believe you still need this last line, otherwise the output variable is not created

MichaelKora: oh good catch...it should be fine..now..also test in PG

actions/changelog-generate/action.yaml
```diff
@@ -73,12 +72,14 @@ runs:
             setfrom=$(git rev-list --max-parents=0 HEAD)
           fi
         fi
+
         echo "fromtag=$setfrom">> $GITHUB_OUTPUT
-        if [ -n "${{ inputs.config }}" ]; then
-          path="$GITHUB_ACTION_PATH/changelog-config.json)"
+        if [ -Z "${{ inputs.config }}" ]; then
```
yannick-roeder, line 77: Did you test this? When I test it locally with `-Z` instead of `-z` then I get the following error:

```
bash: conditional binary operator expected
```

```suggestion
        if [[ -z "${{ inputs.config }}" ]]; then
```

yannick-roeder (changes requested):

yannick-roeder (dismissed):

.github/workflows/python-poetry-release.yaml
MichaelKora: done

yannick-roeder (approved):

disrupted (approved):

PR #156, opened 2023-10-18
Fix Java Gradle Publish action

yannick-roeder:


MichaelKora (approved):

actions/java-gradle-publish/action.yaml
```diff
@@ -54,6 +54,7 @@ runs:
         gradle-cache: ${{ inputs.gradle-cache }}
 
     - name: Publish to Nexus
```
philipp94831, line 56: Is this snapshot publication?

yannick-roeder: Yes, should I name it differently?

philipp94831: Would be helpful

philipp94831 (approved):

PR #159, opened 2023-11-06
Avoid concurrency by canceling triggered worklfows before running new ones

### LINKED ISSUE #158: Add concurrency to our workflows
Ideally the previous workflow is cancelled when the newer workflow is already running. This can be achieved with [workflow concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency). Adding this to every workflow should be sufficient:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-docker
  cancel-in-progress: ${{ github.ref_name != github.event.repository.default_branch }}
```

Note that the suffix of the group is always specific to the workflow to avoid that different workflows cancel each other in the same run.

MichaelKora:
Fixes #158

### DIFF
--- a/.github/workflows/bump-version-release.yaml
+++ b/.github/workflows/bump-version-release.yaml
@@ -44,10 +44,6 @@ on:
         description: "The old version in your `.bumpversion.cfg` file."
         value: ${{ jobs.release.outputs.old-version }}
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-bump-version-release
-  cancel-in-progress: true
-
 jobs:
   release:
     runs-on: ubuntu-22.04
--- a/.github/workflows/bump-version-release.yaml
+++ b/.github/workflows/bump-version-release.yaml
@@ -47,6 +47,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-22.04
+    
     # Map the job outputs to step outputs
     outputs:
       release-version: ${{ steps.bump-version.outputs.release-version }}
--- a/.github/workflows/bump-version-release.yaml
+++ b/.github/workflows/bump-version-release.yaml
@@ -35,9 +35,6 @@ on:
         description: "The GitHub token for committing the changes."
         required: true
 
-    concurrency:
-      group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-bump-version-release
-      cancel-in-progress: true
     # Map the workflow outputs to job outputs
     outputs:
       release-version:
@@ -46,6 +43,9 @@ on:
       old-version:
         description: "The old version in your `.bumpversion.cfg` file."
         value: ${{ jobs.release.outputs.old-version }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-bump-version-release
+  cancel-in-progress: true
 
 jobs:
   release:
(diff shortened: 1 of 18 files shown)


.github/workflows/bump-version-release.yaml
```diff
@@ -44,10 +44,13 @@ on:
         description: "The old version in your `.bumpversion.cfg` file."
         value: ${{ jobs.release.outputs.old-version }}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-bump-version-release
+  cancel-in-progress: true
+
```
yannick-roeder, line 50: We do not want workflows that create tags or releases to every be canceled while they are running

.github/workflows/docker-build-and-publish.yaml
```diff
@@ -67,11 +68,14 @@ on:
         description: "GitHub token."
         required: false
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-docker
+  cancel-in-progress: true
```
yannick-roeder, line 73: Why did you change the `cancel-in-progress` everywhere? The idea is to only cancel the workflow when we are not on the default branch, so that we still let every pipeline run for every merge to main (even if they are running at the same time)

.github/workflows/java-gradle-release.yaml
```diff
@@ -55,6 +56,10 @@ on:
         description: "Bumped version of your project."
         value: ${{ jobs.release.outputs.release-version }}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-java-gradle-release
+  cancel-in-progress: true
+
```
yannick-roeder, line 62: We do not want workflows that create tags or releases to every be canceled while they are running

.github/workflows/helm-gke-deploy.yaml
```diff
@@ -70,11 +71,14 @@ on:
         description: "GKE cluster for authentication"
         required: true
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-helm-deploy
```
yannick-roeder, line 75: For the Kustomize you included the `gke` in the suffix, but for Helm you don't? Please be consistent about your suffixes

.github/workflows/python-poetry-release.yaml
```diff
@@ -57,6 +58,10 @@ on:
         description: "The old version of the package."
         value: ${{ jobs.create-release.outputs.old-version }}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-python-poetry-release
+  cancel-in-progress: true
+
```
yannick-roeder, line 64: We do not want workflows that create tags or releases to every be canceled while they are running

.github/workflows/release-tag-versions.yaml
```diff
@@ -31,10 +32,13 @@ on:
         description: "The GitHub token for committing the changes."
         required: true
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-release-version
+  cancel-in-progress: true
+
```
yannick-roeder, line 38: We do not want workflows that create tags or releases to every be canceled while they are running

yannick-roeder (changes requested):

.github/workflows/helm-gke-deploy.yaml
MichaelKora: my bas.. i missed those two. But the idea is to give the nave of the workflow

yannick-roeder (approved):

PR #175, opened 2024-01-29
Add jib-uploader action

irux:


actions/jib-uploader/action.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build Docker image
+      run: |
+        ./gradlew jib \
+          --info --stacktrace \
+          --image=${{ inputs.image }} \
+          -Djib.container.mainClass=${{ inputs.class }} \
```
raminqaf, line 26: Is this really needs to be required?
It think the user can configure this in its own application.

raminqaf (changes requested):

actions/jib-uploader/action.yaml
yannick-roeder, line 1: What do you think about `java-gradle-build-push-jib` as the action name, so that it can easier be found when looking for Java Gradle actions?

yannick-roeder (changes requested):

actions/jib-uploader/action.yaml
irux: I am not sure to be honest, should I remove it ?

irux: should we still support it as an optional feature or should we remove it completely? what do you guys think ?

raminqaf: I would have it removed. The user should take of the mainClass 🙂

philipp94831 (changes requested):

actions/jib-uploader/action.yaml
philipp94831: We could make it optional as well if there are multiple jib images to be built with different main classes. But I don't have such a use case so I would be fine with removing it

actions/java-gradle-build-push-jib/action.yaml
```diff
@@ -0,0 +1,29 @@
...
+inputs:
+  image:
+    description: "The image name with its repository name"
+    required: true
+  tags:
+    description: "The tags to use to release the image"
+    required: true
+  class:
+    description: "The entrypoint class to be used for the image"
+    required: true
```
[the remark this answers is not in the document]
philipp94831: See discussion above

actions/jib-uploader/action.yaml
irux: @philipp94831 following up this. But how would it work when you have multiple entry points in every module of your project ? do we have such thing atm ? or it is not something it happens quite often ? if not, then I would just remove it.

philipp94831: User needs to configure mainClass in gradle file

irux: Ok then I will remove it

philipp94831:
Implemented by #193

PR #182, opened 2024-02-27
Write build and test cache for all branches

philipp94831:


actions/java-gradle-assess-code-quality/action.yaml
```diff
@@ -10,6 +10,14 @@ inputs:
     description: "Whether Gradle caching is enabled or not. (Default is true)"
     required: false
     default: true
+  gradle-cache-read-only:
+    description: "Whether Gradle caching should be read-only. By default this value is 'false' for workflows on the GitHub default branch and 'true' for workflows on other branches."
+    required: false
+    default: ${{ github.event.repository != null && github.ref_name != github.event.repository.default_branch }}
```
philipp94831, line 16: Same as https://github.com/gradle/actions/blob/main/setup-gradle/action.yml#L21

.github/workflows/java-gradle-docker.yaml
```diff
@@ -97,12 +109,14 @@ jobs:
...
     steps:
       - name: Build tarball image
-        uses: bakdata/ci-templates/actions/java-gradle-build-jib@v1.16.0
+        uses: bakdata/ci-templates/actions/java-gradle-build-jib@feature/improved-gradle-caching
         with:
           java-distribution: ${{ inputs.java-distribution }}
           java-version: ${{ inputs.java-version }}
           gradle-version: ${{ inputs.gradle-version }}
           gradle-cache: ${{ inputs.gradle-cache }}
+          gradle-cache-read-only: true
```
torbsto, line 118: I get that we don't want to write the cache for the test step but why not here and in publish?

philipp94831: For publish I was unsure if we want to write the cache because I don't fully trust the Nexus plugin. For Jib build we could do it I guess. What do you think?

torbsto: Probably it is safer to just disable it, the layers are cached by the docker registry, so I don't think we would gain a lot. But I think we should then include that `gradle-cache-read-only`  is only used for the build + test steps in the description

philipp94831: Added it to the docs

torbsto (dismissed):

yannick-roeder (approved):

PR #188, opened 2024-03-14
Refactor Docker workflow to build and push in a single job

### LINKED ISSUE #120: Reflect branch's latest commit SHA in non-release image tags
The docker-publish action uses variable `GITHUB_SHA` to tag images:
https://github.com/bakdata/ci-templates/blob/8e5f4dbd8ede7c49f1dab22e0c0a6a847e4aa26d/actions/docker-publish/action.yaml#L20

In case the event triggering the docker-publish action is a pull request, the `GITHUB_SHA` used in image tag is a commit SHA that is dynamically generated by GitHub and differs from the branch's latest commit SHA. More information here: https://github.com/orgs/community/discussions/25191#discussioncomment-3246770

To easily match dev deployments to open PRs, it would be helpful if the SHA in the image tag would reflect the branch's latest commit SHA.

### LINKED ISSUE #139: Replace docker-build action in all workflows with official action
## Description

Initially we decided against the official [docker-build-push action](https://github.com/docker/build-push-action) since it didn't seem to support building without pushing. Now there is an input `push` which can be set to false, so we can use this action for building and still rely on our [custom tagging logic](https://github.com/bakdata/ci-templates/blob/main/actions/docker-push/action.yaml#L43-L50) for pushing images build from Dockerfiles and Jib. We want to remove our docker-build action and replace it in every workflow with the [docker-build-push action](https://github.com/docker/build-push-action), because it offers more features.

## Acceptance criteria

- [ ] The `docker-build` action should be removed
- [ ] The docker-build and publish action should be used in every workflow

yannick-roeder:
Fixes #139 
Fixes #120


.github/workflows/docker-build-and-publish.yaml
```diff
@@ -62,57 +62,64 @@ on:
...
+        uses: docker/build-push-action@v5
         with:
-          docker-registry: "${{ inputs.docker-registry }}"
-          image-namespace: "${{ inputs.image-namespace }}"
-          image-name: "${{ inputs.image-name }}"
-          image-tag: "${{ inputs.image-tag }}"
-          image-artifact-name: "${{ inputs.image-artifact-name }}"
-          working-directory: "${{ inputs.working-directory }}"
+          context: ${{ inputs.docker-context }}
+          file: ${{ inputs.docker-context }}/${{ inputs.dockerfile-path }}
```
JakobEdding, line 120: Is it a valid assumption that the Dockerfile always resides inside the Docker context? I think there can be cases where this is not the case and it would be better not to repeat `inputs.docker-context` here.

yannick-roeder: That's the default from the Docker CLI, not sure why it is defined like this:
```
$ docker build --help
Usage:  docker buildx build [OPTIONS] PATH | URL | -

...

Options:
      ...
  -f, --file string                   Name of the Dockerfile (default: "PATH/Dockerfile")
     ...
```

Should I set the default of the `inputs.dockerfile.path` to `./Dockerfile` instead? This would allow more flexibility, but results in the user having to be aware, that the `file:` is not working like it would locally

yannick-roeder: Sadly defaults of inputs don't allow references to other inputs like this:
```
 dockerfile-path:
        description: "Path to the Dockerfile."
        required: false
        default: "${{ inputs.docker-context }}/Dockerfile"
        type: string
```

JakobEdding: It's too bad that we can't set `dockerfile-path.default = null` or something for the inputs of this workflow and let the action set the value from its defined default which is `{context}/Dockerfile`... 🤔 

> Should I set the default of the inputs.dockerfile-path to ./Dockerfile instead?
I would prefer this and then setting `file: ${{ inputs.dockerfile-path }}` here as I think it's less confusing. 

(Also, while the Docker CLI uses `PATH/Dockerfile` as _default_, there you can still override it. Here, as it is currently, you cannot set `file` indepdently of the `docker-context` input.)

yannick-roeder: I set the default of `dockerfile-path` to `./Dockerfile` now

.github/workflows/docker-build-and-publish.yaml
```diff
@@ -62,57 +62,66 @@ on:
...
 jobs:
-  docker-build:
-    name: Docker build
+  build-and-publish:
+    name: Build and Publish
     runs-on: ubuntu-22.04
     steps:
       - name: Check out repository
-        uses: bakdata/ci-templates/actions/checkout@1.32.0
+        uses: bakdata/ci-templates/actions/checkout@feat/refactor-docker
```
yannick-roeder, line 76: Todo

JakobEdding (approved):

PR #201, opened 2024-06-14
Add metadata action to jib push / Merge build+push

DerTiedemann:
I added the metadata action that will create image tags based on the repository config, and will also push those tags to the respective registries. It is assumes that the present docker daemon is current authenticated against the registry you want to push to. This is usually done docker-login action.


actions/java-gradle-build-push-jib-multi-plaftorm/action.yaml
```diff
@@ -69,28 +58,48 @@ runs:
...
-          -DsendCredentialsOverHttp \
-          -Djib.allowInsecureRegistries=true \
-          -Djib.to.image=${{ inputs.full-image-name }}:${{ inputs.image-tag }} \
-          -Djib.to.tags=${{ inputs.image-tag }} \
-          -Djib.to.auth.username=AWS \
-          -Djib.to.auth.password=${{ inputs.registry-password }} \
-          -x test \
+          ${{ inputs.jib-from-image && format('-Djib.from.image={0}', inputs.jib-from-image) || '' }} \
+          ${{ inputs.class && format('-Djib.container.mainClass={0}', inputs.class) || '' }} \
+          -Djib.to.image=${{ inputs.full-image-name }} \
```
yordanovsstoyan, line 101: Please add `-Djib.allowInsecureRegistries=true` and overwrite the original [java-gradle-build-jib](https://github.com/bakdata/ci-templates/tree/main/actions/java-gradle-build-jib) action as this is what Philipp requested. Otherwise action is tested and it works. Only needs the insecure registries flag to be added and `packages: write` to be added under permissions in the workflow that calls the action, so jib can use the docker config

yordanovsstoyan (approved):

PR #216, opened 2024-11-11
Update Python versions of Poetry setup test

yannick-roeder:


DerTiedemann (approved):
LGTM

.github/workflows/_test-python-setup-poetry.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
JakobEdding, line 25: This workflow is only used internally here in this repo, right? Else it might be better to either make the versions here customizable or remove `3.13` because it has only been stable for a short while and not all commonly used packages can be expected to be compatible yet.

yannick-roeder: Yes, it is only used internally in this repo, this is also why I prefixed the workflow with an underscore `_` now.

PR #224, opened 2024-11-26
Fix broken outputs from GSM actions

DerTiedemann:
Unbenannt
As you can see by the commit history - this was a ride which ultimately led me back to the `toJSON` workaround, in which usability suffers quite a bit. There has to be a better solution to this than just chaining actions together and glueing them togehter, but for now this is functional. 

Here is an example workflow that works:
```yaml
name: test

on:
  workflow_dispatch:

permissions:
  contents: read
  id-token: write

jobs:
  test-secret-fetch:
    name: load secrets
    runs-on: ubuntu-latest

    steps:
      - name: Load secrets
        id: load-secrets
        uses: bakdata/ci-templates/actions/gcp-gsm-load-secrets@tiedemann/gsm-object-outputs-fix
        with:
          gke-project-name: <redacted>
          workload-identity-provider: <redacted>
          gke-service-account: <redacted>
          secrets-to-inject: |-
             CRABBY_PATTY_FORMULA/1
      - name: test exported
        run: |
          echo "${{ fromJSON(steps.load-secrets.outputs.secret-json-string).CRABBY_PATTY_FORMULA }}" | rev
```

As you can see this is less then ideal but it works.
Just as a though, having 2 vanilla actions (gcp auth + gsm get secrets) might be a whole lot easier than rolling our own stuff. Maybe we just take the learnings and have sth simple.

### DIFF
--- a/actions/gcp-gsm-load-secrets/action.yaml
+++ b/actions/gcp-gsm-load-secrets/action.yaml
@@ -31,7 +31,7 @@ runs:
         workload_identity_provider: ${{ inputs.workload-identity-provider }}
         service_account: ${{ inputs.gke-service-account }}
     - id: "parse_secrets"
-      uses: "bakdata/ci-templates/actions/parse-secrets-definitions@1.49.0"
+      uses: "bakdata/ci-templates/actions/gcp-gsm-parse-secerts@1.49.0"
       with:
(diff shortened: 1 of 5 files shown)


DerTiedemann:
Superceded in favor of loading the secrets directly using the GSM action

PR #226, opened 2024-11-29
Tag Jib images with branch name and PR # where appropriate

JakobEdding:


philipp94831 (approved):

yannick-roeder (approved):

actions/java-gradle-build-jib/action.yaml
```diff
@@ -89,6 +89,8 @@ runs:
         tags: |
           event=tag,type=semver,pattern={{version}}
           event=push,type=raw,value=pipeline-${{ github.run_id }}-git-{{sha}},enable=${{ github.ref_type != 'tag' }}
+          event=pr,type=ref
```
philipp94831, line 92: As far as I understood, this does not work because the way we inject it into jib, only one tag can be used. I therefore suggest to change our jib usage a bit. Always use the pipeline tag as parameter for `jibImage.tag` and pass additional tags created by metadata action to `jib.to.tags` (git tag, latest, pr)

philipp94831:
Please check if #247 supersedes this now

JakobEdding:
@philipp94831 yes, perfect!

philipp94831:
> @philipp94831 yes, perfect!

@JakobEdding I just saw that the branch tag is not added currently. So if this is needed, feel free to re-open this PR

PR #242, opened 2025-01-15
feat(bump-version): allow dirty working directory by default

disrupted:


philipp94831:
Can you elaborate why this is a good idea? I know our Gradle plugin also enforces a clean git repo

disrupted:
> Can you elaborate why this is a good idea? I know our Gradle plugin also enforces a clean git repo

afaik this is widely used in some of our internal CI templates. Some projects require 2 subsequent bumps. The first one bumps to a snapshot version and the second to the actual release (removing the `-dev` suffix). Currently this requires the workaround of passing `--allow-dirty` in the `release-type` input.
https://github.com/bakdata/kpops/blob/59765dce687a513c680a792e148588936661b0d7/.github/workflows/python-uv-release.yaml#L95

Overall I don't have a strong opinion about this. There are other solutions such as bumping to dev in an earlier step, i.e. `prepare development` commit, like we do in other projects. This is just how it currently works for our Python projects.

DerTiedemann (approved):

PR #245, opened 2025-01-20
Update Gradle actions for new plugin versions

philipp94831:


actions/java-gradle-publish/action.yaml
```diff
@@ -47,7 +47,6 @@ inputs:
     default: "."
 runs:
   using: "composite"
-  if: "!contains(github.repository, 'playground')"
```
philipp94831, line 50: This is a syntax failure. As discussed before, we remove it for now and will fix it later

JakobEdding: Shouldn't it just be `if: ${{ !contains(github.repository, 'playground') }}`

philipp94831: I don't know and I also don't want to fix it as part of this PR. I think the error was that `if` is unexpected but can't find it right now. I also removed it in another PR for another action previously as discussed with the team

philipp94831: https://github.com/bakdata/kafka-large-message-serde/actions/runs/12904617870/job/35982276420

JakobEdding (approved):

actions/java-gradle-publish/action.yaml
JakobEdding: Perhaps open an issue about it? You can either way resolve this convo imo :)

PR #249, opened 2025-01-30
Feature: Cache poetry installation

tobbber:
install poetry runs quite often (e.g. during kpops setup) and always takes 5-10sec. Therefore caching it probably makes sense

### DIFF
--- a/actions/python-setup-poetry/action.yaml
+++ b/actions/python-setup-poetry/action.yaml
@@ -39,7 +39,7 @@ runs:
         path: |
           /opt/pipx/venvs/poetry
           /opt/pipx_bin/poetry
-        key: poetry-${{ inputs.poetry-versions }}-${{ runner.os }}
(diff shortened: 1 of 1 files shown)


yannick-roeder (dismissed):

actions/python-setup-poetry/action.yaml
```diff
@@ -32,7 +32,17 @@ runs:
...
       shell: bash
 
+    - name: Cache poetry
+      id: cache-poetry
+      uses: actions/cache@v4
+      with:
+        path: |
+          /opt/pipx/venvs/poetry
+          /opt/pipx_bin/poetry
+        key: poetry-${{ inputs.poetry-version }}-${{ runner.os }}
```
JakobEdding, line 42: We should probably include `python-version` in the cache key

tobbber: done

PR #256, opened 2025-03-03
Allow configuration of GitHub dependency graph in Gradle

philipp94831:


actions/java-gradle-setup/action.yaml
```diff
@@ -22,6 +22,10 @@ inputs:
     description: "Whether Gradle caching should be read-only. By default this value is 'false' for workflows on the GitHub default branch and 'true' for workflows on other branches."
     required: false
     default: ${{ github.event.repository != null && github.ref_name != github.event.repository.default_branch }}
+  gradle-dependency-graph:
+    description: "Configure GitHub dependency graph for Gradle. See https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#github-dependency-graph-support. (Default is disabled)"
+    required: false
+    default: "disabled"
```
jkbe, line 28: IMO, the more configuration options we add an action like this `java-gradle-setup` action, the weaker the abstraction of the action gets. 

We should think about getting rid of custom actions like this that only add an indirection and obfuscate what's going on with the underlying actions, here `actions/setup-java` and `gradle/actions/setup-gradle`. Just something to keep in mind, I also recently talked to Yannick about this.

jkbe (approved):

PR #261, opened 2025-04-08
feat: create Python uv release workflow

disrupted:


.github/workflows/python-uv-release.yaml
```diff
@@ -0,0 +1,127 @@
...
+          changelog-file: ${{ inputs.changelog-file }}
+
+      - name: Commit and push changes including .bumpversion.cfg file
+        uses: bakdata/ci-templates/actions/commit-and-push@1.25.2
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          commit-message: "Bump version ${{ steps.bump-version-dev.outputs.old-version }} → ${{ steps.bump-version-release.outputs.release-version }}"
+          github-username: ${{ secrets.github-username }}
+          github-email: ${{ secrets.github-email }}
+          github-token: ${{ secrets.github-token }}
```
yannick-roeder, line 117: Don't you need `add-untracked: "true"` here. See https://github.com/bakdata/ci-templates/blob/1.25.2/actions/commit-and-push/action.yaml#L21-L24

.github/workflows/python-uv-release.yaml
```diff
@@ -0,0 +1,127 @@
...
+      - name: Check out repository
+        uses: bakdata/ci-templates/actions/checkout@1.49.0
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false # required for pushing to protected branch later
+          fetch-depth: 0 # required for changelog generation
+
+      - name: Bump version
+        id: bump-version-dev
+        uses: bakdata/ci-templates/actions/bump-version@v1.21.2
```
yannick-roeder, line 86: Consider bumping all the action versions to the newest version if you are creating a new workflow anyway

yannick-roeder (changes requested):

.github/workflows/python-uv-release.yaml
disrupted: good point! added

PR #262, opened 2025-04-14
Skip Sonarqube and Sonatype in private repos

jkbe:


philipp94831 (changes requested):
Please also adapt the library and plugin workflow

philipp94831 (approved):

PR #273, opened 2025-07-23
hotfix 2: also remove git-cliff binary

DerTiedemann:
- **hotfix: remove everything git-cliff related (brutally)**
- **hotfix: also remove git-cliff binary (cold-blooded)**

amend to #272

### DIFF
--- a/actions/changelog-generate/action.yaml
+++ b/actions/changelog-generate/action.yaml
@@ -69,5 +69,5 @@ runs:
         GITHUB_REPO: ${{ github.repository }}
         GITHUB_TOKEN: ${{ inputs.github-token }}
     - name: remove tmp config file
-      run: rm tmp_cliff.toml && rm -rf git-cliff*
+      run: rm tmp_cliff.toml && rm -rf git-cliff* && rm -rf bin/git-cliff
       shell: bash
--- a/actions/changelog-generate/action.yaml
+++ b/actions/changelog-generate/action.yaml
@@ -58,8 +58,7 @@ runs:
         fi
       shell: bash
     - name: Generate latest only changelog
-      # there is no proper release for the install dir fix, yet
-      uses: orhun/git-cliff-action@104a6cf3c9aa0fdfe4eab129f9c1900e1eb8f7fd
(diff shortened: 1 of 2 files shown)


actions/changelog-generate/action.yaml
```diff
@@ -50,24 +58,30 @@ runs:
         fi
       shell: bash
     - name: Generate latest only changelog
-      uses: orhun/git-cliff-action@v4
+      # there is no proper release for the install dir fix, yet
+      uses: orhun/git-cliff-action@104a6cf3c9aa0fdfe4eab129f9c1900e1eb8f7fd
```
jkbe, line 62: This was released now in https://github.com/orhun/git-cliff-action/releases/tag/v4.5.0 and is therefore also part of `v4`.

Same for the `git-cliff-action` usage below.

actions/changelog-generate/action.yaml
```diff
@@ -50,24 +58,28 @@ runs:
...
       id: git-cliff-latest
       with:
         config: tmp_cliff.toml
-        args: "--unreleased --tag ${{ inputs.tag }} --verbose"
+        args: "--unreleased --tag ${{ inputs.tag }} ${{ inputs.verbose == 'true' && '--verbose' || '' }} ${{ inputs.additional-cliff-args }}"
       env:
+        # I am not a fan of this, cuz this does not reflect what this step is supposed to do
+        # i.e. just dumping the latest changelog into GHA context
+        # BUT, it works and the file will be written over in the next step anyway
+        OUTPUT: ${{ inputs.changelog-file }}
```
jkbe, line 70: Is this still necessary? It seems like there is a default file that git-cliff writes to: https://github.com/orhun/git-cliff-action/blob/98c93442bb05a455a77bee982867857ae748eeea/run.sh#L21

yannick-roeder (approved):

actions/changelog-generate/action.yaml
DerTiedemann: that file would have to be explicitly removed after creation, otherwise it would end up in the diff.
I dont really mind adding sth to the rm statement in the cleanup step

PR #280, opened 2025-08-07
fix: let Poetry assume custom Python version in setup action

DerTiedemann:
I have no idea how this worked previously

### DIFF
--- a/actions/python-setup-poetry/action.yaml
+++ b/actions/python-setup-poetry/action.yaml
@@ -22,6 +22,9 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version }}
     - name: Install Poetry ${{ inputs.poetry-version }}
       shell: bash
       run: pipx install poetry==${{ inputs.poetry-version }}
--- a/actions/python-setup-poetry/action.yaml
+++ b/actions/python-setup-poetry/action.yaml
@@ -49,7 +49,7 @@ runs:
     - name: Install Poetry ${{ inputs.poetry-version }}
       if: steps.cache-poetry.outputs.cache-hit != 'true'
       shell: bash
-      run: pipx install poetry==${{ inputs.poetry-version }} --python ${{ steps.setup-python.outputs.python-path }}
+      run: pipx install poetry==${{ inputs.poetry-version }} --python "${{ steps.setup-python.outputs.python-path }}"
 
     - name: Set up Python ${{ inputs.python-version }} poetry cache
       uses: actions/setup-python@v5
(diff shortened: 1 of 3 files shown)


actions/python-setup-poetry/action.yaml
```diff
@@ -43,23 +48,18 @@ runs:
...
-      run: |
-        poetry config virtualenvs.in-project true
-        poetry config virtualenvs.path ${{ inputs.working-directory }}/.venv
-      shell: bash
-
-    - name: Set up Python ${{ inputs.python-version }}
-      uses: actions/setup-python@v4
-      id: setup-python
+    - name: Set up Python ${{ inputs.python-version }} poetry cache
+      uses: actions/setup-python@v5
```
jkbe, line 55: Have you tried to merely pull up this `setup-python` step to the top instead of introducing a 2nd one at the top? Perhaps that's already enough to fix this.

DerTiedemann: I tried that. It failed requiring the poetry binary.

jkbe: How do the `setup-python` steps relate to Poetry being available? That happens through `pipx install`. If that was really the issue, it sounds a bit like an issue with GHA cache?

actions/python-setup-poetry/action.yaml
```diff
@@ -43,23 +48,18 @@ runs:
...
       with:
         python-version: ${{ inputs.python-version }}
-        cache: poetry
+        cache: "poetry"
         cache-dependency-path: ${{ inputs.working-directory }}/poetry.lock
 
-    - run: poetry env use "${{ steps.setup-python.outputs.python-path }}"
```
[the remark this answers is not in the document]
jkbe: But you did remove it. Hence I'm asking.

actions/python-setup-poetry/action.yaml
DerTiedemann: https://github.com/bakdata/github-playground-2/actions/runs/16801695033/job/47584394585
this is the correct run that points out that the binary is misisng

actions/python-setup-poetry/action.yaml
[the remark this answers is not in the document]
DerTiedemann: yeah, brainfart - i removed it as i assumed with poetry being directly installed to the proper python install i would not need this.

actions/python-setup-poetry/action.yaml
jkbe: Yes I get it now that it's due to the poetry dependency caching that's part of `setup-python`

jkbe (changes requested):
Please prebump usage of the action in [.github/workflows/python-poetry-publish-pypi.yaml](https://github.com/bakdata/ci-templates/blob/c92dc39dc970199fd1aadc15a550bdac582d6aba/.github/workflows/python-poetry-publish-pypi.yaml#L45) and [actions/python-poetry-bump-version/action.yaml](https://github.com/bakdata/ci-templates/blob/c92dc39dc970199fd1aadc15a550bdac582d6aba/actions/python-poetry-bump-version/action.yaml#L33). LGTM otherwise!

yannick-roeder (approved):

PR #284, opened 2025-10-16
tiedemann/debloat action and docker workflow integration

DerTiedemann:
- **feat: add debloat action**
- **feat: integrate debloat action into docker workflow behind a flag**


actions/debloat-runner/action.yaml
```diff
@@ -0,0 +1,13 @@
+name: "Debloat runner"
+description: "Remove unnessecary libraries & binaries for more leeway on space intensive builds"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout repository code
+      uses: actions/checkout@v4
```
yannick-roeder: Is checking out here necessary? I think it might even overwrite some authorization towards GitHub in the worst case since `persist-credentials: false` is not set.

yannick-roeder (changes requested):

yannick-roeder (dismissed):
Please test again, but then it is good to be merged

PR #309, opened 2026-04-17
fix: Use full Python semver in key to fix `python-setup-poetry` cache

jkbe:
Fixes https://github.com/bakdata/ci-templates/issues/286

### DIFF
--- a/.github/workflows/python-poetry-publish-pypi.yaml
+++ b/.github/workflows/python-poetry-publish-pypi.yaml
@@ -42,7 +42,7 @@ jobs:
         uses: bakdata/ci-templates/actions/checkout@1.81.1
 
       - name: Set up Poetry ${{ inputs.poetry-version }}
-        uses: bakdata/ci-templates/actions/python-setup-poetry@1.81.1
+        uses: bakdata/ci-templates/actions/python-setup-poetry@fix/poetry-binary-cache
         with:
           python-version: ${{ inputs.python-version }}
           poetry-version: ${{ inputs.poetry-version }}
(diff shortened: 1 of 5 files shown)


actions/python-setup-poetry/action.yaml
```diff
@@ -37,18 +37,17 @@ runs:
...
-#          /opt/pipx_bin/poetry
-#        key: poetry${{ inputs.poetry-version }}-python${{ inputs.python-version }}-${{ runner.os }}
+    - name: Cache Poetry binary
+      id: cache-poetry
+      uses: actions/cache@v5
+      with:
+        path: |
+          /opt/pipx/venvs/poetry
+          /opt/pipx_bin/poetry
+        key: pipx-poetry-${{ inputs.poetry-version }}-python-${{ steps.setup-python.outputs.python-version }}-${{ runner.os }}
```
jkbe, line 47: It seems that the Poetry binary starts with a shebang that includes the full 3-part Python version. The `inputs.python-version` is typically only a 2-part version like `3.13`. Therefore, the cached Poetry broke as soon as `setup-python` above installed a newer Python patch version and the old cache still had a Poetry binary with the older Python patch version in its shebang

philipp94831 (approved):

.github/workflows/python-poetry-publish-pypi.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         uses: bakdata/ci-templates/actions/checkout@1.81.1
 
       - name: Set up Poetry ${{ inputs.poetry-version }}
-        uses: bakdata/ci-templates/actions/python-setup-poetry@1.81.1
+        uses: bakdata/ci-templates/actions/python-setup-poetry@fix/poetry-binary-cache
```
philipp94831: Bump

philipp94831 (changes requested):

philipp94831 (approved):

